PR #31, opened 2014-06-21
use a standard string instead of PATH

brad:
When I was setting this up, I found that the example code didn't work because of this undefined `PATH`. Just using a standard string worked fine.


hkliya:
This should be closed?

rougeth:
`PATH` is not being imported on `README.md`, so it's not clear if it is a real function or if it is just alerting that specific field should be a path to `.apk`. I believe we should change the example to use absolute path, it would make it clear to understand.

PR #68, opened 2015-01-08
Add base class for Sauce tests

imurchie:
So there is a simple way to use the client with Sauce.

### DIFF
--- a/appium/__init__.py
+++ b/appium/__init__.py
@@ -15,3 +15,5 @@
 """
 Appium Python Client
 """
+from .saucetestcase import SauceTestCase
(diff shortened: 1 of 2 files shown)


jlipps:
:+1:

rgonalo:
Hi Isaac, 
After this commit, appium tests fail if user has not installed 'sauceclient'.
To avoid installing an optional library, maybe it would be better not to import SauceTestCase in __init__.py or catch import exception. What do you think?

imurchie:
Yes. My apologies. I will fix this. Thanks!

PR #71, opened 2015-02-16
Implemented set_location() - gps spoofing

jwallis:
iOS: Sometimes the call hangs if you are using an app that is requesting/using location (a UIAutomaion issue)
Android: I could not test because I cannot get the emulator working on my machine, but it should work

### DIFF
--- a/appium/webdriver/mobilecommand.py
+++ b/appium/webdriver/mobilecommand.py
@@ -54,3 +54,4 @@ class MobileCommand(object):
     START_ACTIVITY = 'startActivity'
     GET_SETTINGS = 'getSettings'
     UPDATE_SETTINGS = 'updateSettings'
+    SET_LOCATION = 'setLocation'
--- a/appium/webdriver/webdriver.py
+++ b/appium/webdriver/webdriver.py
@@ -693,6 +693,18 @@ class WebDriver(webdriver.Remote):
         self.execute(Command.TOGGLE_LOCATION_SERVICES, {})
         return self
 
+    def set_location(self, latitude, longitude):
+        """Set location for iOS simulator
+
+        :Args:
+         - latitude - String or numeric. Values should be between -90.0 and 90.0
+         - longitude - String or numeric. Values should be between -180.0 and 180.0
+        """
+
+        data = {"latitude": str(latitude), "longitude": str(longitude)}
+
+        self.execute(Command.SET_LOCATION, data)
+        return self
 
     def _addCommands(self):
         self.command_executor._commands[Command.CONTEXTS] = \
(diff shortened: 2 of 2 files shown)


jwallis:
Let me know if I should have done this PR differently

jlipps:
@jwallis isn't `location` already a part of the webdriver client?

imurchie:
The situation is a bit odd. The WebDriver client implements the [command end point](https://code.google.com/p/selenium/source/browse/py/selenium/webdriver/remote/command.py#122) and the [link to the correct point in the JSON Wire Protocol](https://code.google.com/p/selenium/source/browse/py/selenium/webdriver/remote/remote_connection.py#290). But it doesn't actually implement any method that allows one to call it. 

So part of this PR is needed. Not sure if we want to keep it all in the Appium client for now, so it is not confusing.

jlipps:
we should contribute it to the selenium project too

jlipps:
(cc @lukeis do you know why this situation exists?)

jwallis:
I've been generally unclear as to what selenium functionality should be wrapped and what should not.  As @imurchie said, you can use it... if you can discover that it exists.

jwallis:
Ok I think I get it now. Since selenium defines SET_LOCATION and its routing, the method should probably just be moved there.

lukeis:
This doesn't match the current Java bindings:

https://github.com/SeleniumHQ/selenium/blob/master/java/client/src/org/openqa/selenium/remote/html5/RemoteLocationContext.java#L48

Which from how I read it (sorry, didn't test it) would send the data payload as:
    { "location": {"latitude": latitude_value, "longitude": longitude_value, "altitude": altitude_value} }

And yes, this should probably get into the main selenium project, but seems like appium might not have implemented this the same as the existing api?

lukeis:
as far as the why, this is an endpoint thrown together by the googlers and wasn't well defined and since no one was requesting it in python, only the commands were listed.  Allowing one if they were so inclined to do this directly:

    driver.execute(Command.SET_LOCATION, {....})

jlipps:
> but seems like appium might not have implemented this the same as the existing api?

I think we did, we use the built-in location commands in the other selenium clients that have them.

> as far as the why, this is an endpoint thrown together by the googlers and wasn't well defined and since no one was requesting it in python, only the commands were listed. Allowing one if they were so inclined to do this directly

makes sense, we should contribute this to the selenium python client then

imurchie:
I think @lukeis is correct about this PR having the wrong payload. Here we'll get

```json
{
    "latitude": "latitude_value", 
    "longitude": "longitude_value"
}
```

rather than as the [spec](https://code.google.com/p/selenium/wiki/JsonWireProtocol#/session/:sessionId/location) says

```json
{ 
    "location": {
        "latitude": latitude_value, 
        "longitude": longitude_value, 
        "altitude": altitude_value
    }
 }
```

The first ought to produce a deprecation warning in Appium (see [here](https://github.com/appium/appium/blob/master/lib/server/controller.js#L834))

jwallis:
Ugh, you're right.  I did not notice the warning before but see it now, sorry.

saltpy:
I think I have submitted that with the altitude fix in there - I'm hoping this can get fixed and out as I am depending on the functionality.

imurchie:
Closed with #82.

Issue #49, opened 2014-08-22
Create isLocked method


austenke:
https://github.com/appium/appium/pull/3451

KazuCocoa:
https://github.com/appium/python-client/blob/933451122a91e0ea9aafb38aa484e75563e6fa89/appium/webdriver/webdriver.py#L898

PR #84, opened 2015-05-20
Added WebDriver methods to unlock device and check whether the device is currently locked

tylersnavely-wf:
Resolves: https://github.com/appium/python-client/issues/49

### DIFF
--- a/appium/webdriver/webdriver.py
+++ b/appium/webdriver/webdriver.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 from selenium import webdriver
+from selenium.webdriver.support.wait import WebDriverWait
 
 from .connectiontype import ConnectionType
 from .mobilecommand import MobileCommand as Command
@@ -572,7 +573,7 @@ class WebDriver(webdriver.Remote):
         return self.execute(Command.END_TEST_COVERAGE, data)['value']
 
     def lock(self, seconds):
-        """Lock the device for a certain period of time. iOS only.
+        """Lock the device for a certain period of time.
 
         :Args:
          - the duration to lock the device, in seconds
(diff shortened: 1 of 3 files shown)


appium/webdriver/webdriver.py
```diff
@@ -581,8 +582,23 @@ def lock(self, seconds):
             'seconds': seconds,
         }
         self.execute(Command.LOCK, data)
+        if self.desired_capabilities.get('platformName', '').lower() == 'android':
+            WebDriverWait(self, 10).until(lambda d: d.is_locked())
```
tylersnavely-wf: This is really up for discussion.  I personally believe that the Appium server should be responsible for this, not the client.  If the maintainers decide not to add this, that's totally understandable.  It was just helpful for "robustifying" while testing the change.
I'm only running this check on Android because iOS should be back in an unlocked state by the end of the `execute` call.

jlipps: i agree that the server should be responsible for only returning when we're in a known locked state

jlipps: so for now, if it's flakey, we can just show users how to add the wait in their own client code. meanwhile, we'll work on making the appium server more responsible :-)

thomascarpentier:
Hi @jlipps , @tylersnavely-wf 

I'm interested to have this functionality in my python code. Do you have some news? This PR is still alive?

tylersnavely-wf:
Sorry @thomascarpentier , I had honestly forgotten about this.  I really haven't done anything with the project since this was created.  It looks like stripping the additions to `lock()` is all that is needed along with resolving some merge conflicts.  I will try to find some time this week to clean this up.

appium/webdriver/webdriver.py
KazuCocoa: We'd love to remove wait in each method and make users insert wait in their own code.
(Other clients also don't have `wait` in Appium client side.)

mayuayla:
Looking forward to be able to use unlock in Python.  Thanks for working on this.

PR #87, opened 2015-06-04
Add location_in_view property

### LINKED ISSUE #86: Add location_in_view accessor to element
Start with Appium 1.2.x the location_in_view on an element was supported.
We need location_in_view so that people can move away from just using the older 'location' accessor because that will eventually be deprecated.

imurchie:
`WebElement` needs a `location_in_view` method to replace `location`, which is deprecated. Resolves #86.

Also fix a couple of bugs.


jlipps:
:+1: can you take care of a release?

imurchie:
Yes. I'll get on it.

curtissiemens:
@imurchie 
I tried removing webdriver.py line 49 & then everything worked (and set_value was a method on the element)
WebElement.set_value = set_value

Could you please make this change and produce a 0.16 build?

imurchie:
Oh shoot. I'm sorry.

imurchie:
Sorry. That file slipped through when I committed, as I thought it was part of my experimental work in the client. Fixed.

PR #90, opened 2015-06-10
Make tap duration be handled as ms, not s

### LINKED ISSUE #89: The duration argument of WebDriver.tap is inconsistent between code and doc.
I found the doc and code inconsistent in `WebDriver.tap` method.

The doc of `WebDriver.tap` says argument `duration` is in milliseconds:
> \- duration - (optional) length of time to tap, in ms

But the code deal with duration as it's in seconds:

```
if duration:
    duration *= 1000  # we take seconds, but send milliseconds
    action.long_press(x=x, y=y, duration=duration).release()
```

Am I misunderstand anything?

imurchie:
Duration sent into `driver.tap` ought to be `ms`, like the documentation says, not `s`, as the multi-tap case handles it.

Fixes #89

### DIFF
--- a/appium/webdriver/webdriver.py
+++ b/appium/webdriver/webdriver.py
@@ -196,7 +196,6 @@ class WebDriver(webdriver.Remote):
             x = positions[0][0]
             y = positions[0][1]
             if duration:
-                duration = duration
                 action.long_press(x=x, y=y, duration=duration).release()
             else:
                 action.tap(x=x, y=y)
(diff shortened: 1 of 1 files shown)


appium/webdriver/webdriver.py
```diff
@@ -208,7 +208,7 @@ def tap(self, positions, duration=None):
                 y = position[1]
                 action = TouchAction(self)
                 if duration:
-                    duration *= 1000  # we take seconds, but send milliseconds
+                    duration = duration
```
jlipps: anything look funny to you about this line? :-)

imurchie: Not at all! smh.

jlipps:
:+1: will let you release

PR #91, opened 2015-06-11
add wait_activity method for WebDriver

WKPlus:
add an wait_activity method for WebDriver which is helpful to make test case stable without sacrificing efficiency (sleeping a long time does).

### DIFF
--- a/appium/webdriver/webdriver.py
+++ b/appium/webdriver/webdriver.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import time
+
 from selenium import webdriver
 
 from .connectiontype import ConnectionType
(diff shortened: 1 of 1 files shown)


appium/webdriver/webdriver.py
```diff
@@ -400,6 +402,23 @@ def current_activity(self):
         """
         return self.execute(Command.GET_CURRENT_ACTIVITY)['value']
 
+    def wait_activity(self, ac, retry=10, interval=1):
```
imurchie: I would prefer `app_activity` over `ac`, `retries` over `retry`, and something like `timeout` over `interval`, to keep things in line with the naming conventions throughout the module.

appium/webdriver/webdriver.py
```diff
@@ -400,6 +402,23 @@ def current_activity(self):
...
+        """Wait for an activity: block until target activity presents or retry
+        time exhausts.
+
```
imurchie: Should be noted that this is Android only.

imurchie:
A couple of comments. Otherwise, it looks good and useful! Thanks!

appium/webdriver/webdriver.py
WKPlus: how about shorter `activity` instead of `app_activity`? And yes, `retries` is better than `retry`. Will `timeout` make it seems like to be the total time out setting?

appium/webdriver/webdriver.py
WKPlus: Yes, I will add it.

WKPlus:
Do I need to squash two commits into one?

imurchie:
That would be preferable, yes.

WKPlus:
Commits squashed.

imurchie:
Excellent. Thank you for this!

PR #92, opened 2015-06-16
modify the implementation of wait_activity method

WKPlus:
Found a easier way to implement `wait_activity` method: using selenium `WebDriverWait`


imurchie:
Looks good. Thanks!

PR #110, opened 2015-10-24
Update README.md

tophercf:
smallest win in history


imurchie:
Any win is a good win! Thanks!

Issue #112, opened 2015-11-20
Cannot use SauceTestCase under Python 3


Ling:
Module 'new' was removed in Python 3 (deprecated since 2.6) and replaced with types.

python3 -m appium.saucetestcase
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/runpy.py", line 170, in _run_module_as_main
    "__main__", mod_spec)
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/appium/saucetestcase.py", line 18, in <module>
    import new
ImportError: No module named 'new'

imurchie:
Fixed by #113

PR #113, opened 2015-11-20
Fix saucetestcase to run under Python3.

Ling:
The module 'new' was removed. Instead of new.newclass, use type().

This fixed issue #112.


imurchie:
This is great! Thank you!

PR #146, opened 2016-12-15
dontStopAppOnReset instead of stopAppOnReset

z00sts:
Found this change in 1.4
https://github.com/appium/appium/blob/master/CHANGELOG.md#changes-in-version-14-from-137
```
new capability disableAndroidWatchers
deprecated capability stopAppOnReset
```


imurchie:
Thanks!

PR #154, opened 2017-02-06
Added get_page_source command.

benzhou29:
Added get_page_source command.


imurchie:
Does the standard Selenium command not work?

http://selenium-python.readthedocs.io/api.html?highlight=source#selenium.webdriver.remote.webdriver.WebDriver.page_source

benzhou29:
Yes, it works. Didn't realize it existed. Thanks, I will close this then.

PR #157, opened 2017-03-23
Add toggleTouchIdEnrollment

dpgraham:


appium/webdriver/webdriver.py
```diff
@@ -630,14 +630,19 @@ def shake(self):
...
             'match': match
         }
         self.execute(Command.TOUCH_ID, data)
         return self
 
+    def toggle_touch_id_enrollment(self):
+        """Toggle enroll touchId on iOS Simulator
+        """
+        self.execute(Command.TOGGLE_TOUCH_ID_ENROLLMENT)
+
```
imurchie: This should return `self` for chaining's sake.

Issue #168, opened 2017-05-31
driver.swipe() parameter description is inconsistent with appium server


oscheller1:
Tested with iOS 10.3. Not sure whether this is a python client or XCUITest driver problem.
Currently the `driver.swipe(...)` documentation is the following:

```
def swipe(self, start_x, start_y, end_x, end_y, duration=None) Inferred type: (self: WebDriver, start_x: Any, start_y: Any, end_x: Any, end_y: Any, duration: Any) -> WebDriver  
Swipe from one point to another point, for an optional duration.
  
Args:
start_x - x-coordinate at which to start
start_y - y-coordinate at which to start
end_x - x-coordinate at which to stop
end_y - y-coordinate at which to stop
duration - (optional) time to take the swipe, in ms.
```

**Problem**: The appium server is treating the `end_x` and `end_y` parameters as relative values to `start_x` and `start_y` instead of absolute coordinates.

oscheller1:
If it's an accepted problem, here's a PR: https://github.com/appium/python-client/pull/169

jlipps:
@imurchie did we ever decide whether the server or client was doing the right thing?

oscheller1:
Actually this seems to be an inconsistency between Android and iOS. `swipe()` on iOS uses relative  target coordinates, whereas `swipe` on Android uses absolute target coordinates.

oscheller1:
Logs showcasing the inconsistency.

Android: 
```
[HTTP] --> POST /wd/hub/session/ccf9eb5b-fb8a-48cd-bee8-079c0d709ba5/touch/perform {"sessionId":"ccf9eb5b-fb8a-48cd-bee8-079c0d709ba5","actions":[{"action":"press","options":{"x":540,"y":762.5}},{"action":"wait","options":{"ms":2000}},{"action":"moveTo","options":{"x":-262,"y":0}},{"action":"release","options":{}}]}
[debug] [MJSONWP] Calling AppiumDriver.performTouch() with args: [[{"action":"press","options":{"x":540,"y":762.5}},{"action":"wait","options":{"ms":2000}},{"action":"moveTo","options":{"x":-262,"y":0}},{"action":"release","options":{}}],"ccf9eb5b-fb8a-48cd-bee8-079c0d709ba5"]
[debug] [AndroidBootstrap] Sending command to android: {"cmd":"action","action":"swipe","params":{"startX":540,"startY":762.5,"endX":-262,"endY":0,"steps":56}}
```

iOS: 
```
[...]

PR #169, opened 2017-05-31
Fix `swipe` parameter description

oscheller1:
Fix for this issue: https://github.com/appium/python-client/issues/168


oscheller1:
Problem requires review, see https://github.com/appium/python-client/issues/168#issuecomment-315681791

PR #181, opened 2017-08-31
Allow different strategies for hide_keyboard

Ysoko:
Only if key_name, key, and strategy are None do we need to set the strategy to 'tapOutside'. This change allows setting just the strategy to some other value, like 'swipeDown'.


cwaddell:
Thanks for submitting this for me Ysoko!

KazuCocoa:
I've merged this since it has already approved.

PR #224, opened 2018-06-12
add find by image commands and tests

jlipps:
cc @imurchie @dpgraham @mykola-mokhnach @KazuCocoa .

This represents my vision for how clients should expose the template matching functionality with opencv. I.e., return a "clone" of WebElement with a smaller set of methods, so users can imagine they are simply finding "elements" and clicking on them.

Note that tests will only pass with https://github.com/appium/appium-support/pull/74 applied to the Appium server used for testing.


test/functional/android/find_by_image_tests.py
```diff
@@ -0,0 +1,75 @@
...
+        desired_caps = desired_capabilities.get_desired_capabilities('ApiDemos-debug.apk')
+        self.driver = webdriver.Remote('http://localhost:4723/wd/hub', desired_caps)
+
+    def tearDown(self):
+        self.driver.quit()
+
+    def test_find_based_on_image_template(self):
+        image_path = desired_capabilities.PATH('find_by_image_success.png')
+        el = WebDriverWait(self.driver, 3).until(
+            EC.presence_of_element_located((By.IMAGE, image_path))
```
KazuCocoa: ah.. interesting.

appium/webdriver/webdriver.py
```diff
@@ -327,6 +339,43 @@ def find_elements_by_android_uiautomator(self, uia_string):
         """
         return self.find_elements(by=By.ANDROID_UIAUTOMATOR, value=uia_string)
 
+    def find_element_by_image(self, png_img_path):
```
mykola-mokhnach: I think it might be handy to pass the threshold to this method explicitly and set it to 0.5 by default, so clients can customise the behaviour if needed

appium/webdriver/webdriver.py
```diff
@@ -327,6 +341,44 @@ def find_elements_by_android_uiautomator(self, uia_string):
         """
         return self.find_elements(by=By.ANDROID_UIAUTOMATOR, value=uia_string)
 
+    def find_element_by_image(self, png_img_path,
+                              match_threshold=DEFAULT_MATCH_THRESHOLD):
+        """Finds a portion of a screenshot by an image.
+        Uses driver.find_image_occurrence under the hood.
+
+        :Args:
+        - png_img_path - a string corresponding to the path of a PNG image
```
KazuCocoa: Should we address `match_threshold`?

appium/webdriver/webdriver.py
```diff
@@ -327,6 +341,44 @@ def find_elements_by_android_uiautomator(self, uia_string):
...
+        """
+        els = []
+        try:
+            els.append(self.find_element_by_image(png_img_path, match_threshold))
+        except NoSuchElementException:
+            pass
+        return els
```
KazuCocoa: 👍

jlipps:
@imurchie mind publishing a new version when you get a chance?

PR #227, opened 2018-06-20
Set None as default value to lock device

Madh93:
To lock forever until `unlock()` is called:

    device.lock()

To lock for 5 seconds:

    device.lock(5)

### DIFF
--- a/appium/webdriver/webdriver.py
+++ b/appium/webdriver/webdriver.py
@@ -940,11 +940,11 @@ class WebDriver(webdriver.Remote):
         }
         return self.execute(Command.END_TEST_COVERAGE, data)['value']
 
-    def lock(self, seconds):
+    def lock(self, seconds=0):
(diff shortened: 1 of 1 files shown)


appium/webdriver/webdriver.py
```diff
@@ -940,11 +940,11 @@ def end_test_coverage(self, intent, path):
         }
         return self.execute(Command.END_TEST_COVERAGE, data)['value']
 
-    def lock(self, seconds):
+    def lock(self, seconds=0):
```
KazuCocoa: Thanks! `zero` is also good, but could you defien `None` like below? (The behaviour is the same as `zero` or less case.)

```python
def lock(self, seconds=None):
    if seconds is None:
        self.execute(Command.LOCK)
    else:
        self.execute(Command.LOCK, {'seconds': seconds})

return self
```

If you have no time to update, it's fine for current way.

PR #243, opened 2018-08-15
Autopepe

KazuCocoa:
What about adding auto formatter for thire repository following https://www.python.org/dev/peps/pep-0008/ ?

It will help collaborators to fix syntax ambiguous automatically. If we have syntax issue which can fix by autopep8 automatically, `ci.sh` raises an error.

cc @VenkateshPS

### DIFF
--- a/appium/saucetestcase.py
+++ b/appium/saucetestcase.py
@@ -32,7 +32,7 @@ def on_platforms(platforms):
         module = sys.modules[base_class.__module__].__dict__
         for i, platform in enumerate(platforms):
             name = "%s_%s" % (base_class.__name__, i + 1)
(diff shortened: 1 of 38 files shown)


appium/saucetestcase.py
```diff
@@ -48,7 +48,8 @@ def setUp(self):
         self.driver.implicitly_wait(30)
 
     def tearDown(self):
-        print("Link to your job: https://saucelabs.com/jobs/%s" % self.driver.session_id)
+        print("Link to your job: https://saucelabs.com/jobs/%s" %
```
mykola-mokhnach: It makes sense to configure max line length at 120 chars

KazuCocoa:
thanks!

Issue #254, opened 2018-10-03
Not working scroll from one element to another as expected in Appium 1.9.1 using python client with the  robotframework-appiumlibrary on the iOS 12 


WaFeIS7:
## The problem
Maybe it is not the issue, just improvement which needs to do.

To create automation tests on the mobile devices, I use robotframework and robotframework-appiumlibrary, appium python client. 

Since iOS released 12 version, I set up the latest appium 1.9.1 and tried to run my tests on the real iOS 12 device. My tests were broken because the scroll(from one element to the another) didn`t work as expected. 
In the server appium I see response status of the request 200, but nothing happens on the device. 

I asked advice @mykola-mokhnach and his advice helped, he said that need to add wait after the press in the TouchAction

I take on only appium-python and tried to scroll . 
`action = TouchAction(self.driver)
 action.press(element1).wait(600).move_to(element2).release().perform()`.  It works on the iOS 12.
But in the robotframework-appiumlibrary I can't make this function. In the Library there is no such implementation

The function(scroll - which I use in my tests) in the robotframework-appiumlibrary used scroll function from Appium_Python_Client-0.28-py2.7.egg!/appium/webdriver/webdriver.py

> Function scroll in the robotframework-appiumlibrary: 

def scroll(self, start_locator, end_locator):
el1 = self._element_find(start_locator, True, True)
el2 = self._element_find(end_locator, True, True)
driver = self._current_application()
driver.scroll(el1, el2)

> Function scroll in the Appium_Python_Client: 
    

def scroll(self, origin_el, destination_el):
        :Args:
         - originalEl - the element from which to being scrolling
         - destinationEl - the element to scroll to
        :Usage:
            driver.scroll(el1, el2)
        """
        action = TouchAction(self)
        action.press(origin_el).move_to(destination_el).release().perform()
        return self

In this function there is no wait after function press, I think that here there is the issue which could not allow working function scroll correct in the robotframework-appiumlibrary

## Environment
-  Appium version 1.9.1:
-  Xcode 10
-  Appium-Python-Client==0.28
-  robotframework-appiumlibrary==1.4.6
-  Python 2.7
- Node.js version v9.11.1
- iOS 12
- Real device 

## Details
My Local Solution is: 
I took the Appium-Python-Client, found function scroll in the WebDriver Class and add the third argument duration and wait after press: 

>  def scroll(self, origin_el, destination_el, duration):
        action = TouchAction(self)
[...]

PR #256, opened 2018-10-04
add a duration for scroll for ios

KazuCocoa:
Not working scroll from one element to another as expected in Appium 1.9.1 using python client with the robotframework-appiumlibrary on the iOS 12 #254


appium/webdriver/webdriver.py
```diff
@@ -425,18 +425,19 @@ def create_web_element(self, element_id):
         return MobileWebElement(self, element_id)
 
     # convenience method added to Appium (NOT Selenium 3)
-    def scroll(self, origin_el, destination_el):
+    def scroll(self, origin_el, destination_el, duration = 250):
```
mykola-mokhnach: the default duration should be at least 600 ms

Issue #265, opened 2018-10-25
driver.set_clipboard() and driver.set_clipboard_text doesn't work


Mywayking:
## The problem

Briefly describe the issue you are experiencing (or the feature you want to see added to Appium). Tell us what you were trying to do and what happened instead. Remember, this is _not_ a place to ask questions. For that, go to http://discuss.appium.io!

## Environment
- Appium version (or git revision) that exhibits the issue:
- Last Appium version that did not exhibit the issue (if applicable):
- Desktop OS/version used to run Appium:
- Node.js version (unless using Appium.app|exe):
- Mobile platform/version under test:
- Real device or emulator/simulator:
- Appium CLI or Appium.app|exe:

## Details

If necessary, describe the problem you have been experiencing in more detail.

## Link to Appium Logs

Create a [GIST](https://gist.github.com) which is a paste of your _full_ Appium logs, and link them here.

KazuCocoa:
Can you attach the server & client logs?

JeanMaritain:
### driver.set_clipboard_text() not successful
throws: TypeError: Object of type 'bytes' is not JSON serializable
### enviroment
- Appium Version 1.6.3
- Appium-Python-Client 0.29 
- Python 3.6
 ### python client
```call stack
  File "/Users/superuser/PycharmProjects/MobileAutoTest/pageobjects/simplepgs.py", line 68, in set_clipboard
    self.driver.set_clipboard(bytes(content, 'utf8'))
  File "/Users/superuser/PycharmProjects/MobileAutoTest/venv/lib/python3.6/site-packages/appium/webdriver/webdriver.py", line 1279, in set_clipboard
    self.execute(Command.SET_CLIPBOARD, options)
  File "/Users/superuser/PycharmProjects/MobileAutoTest/venv/lib/python3.6/site-packages/selenium/webdriver/remote/webdriver.py", line 319, in execute
    response = self.command_executor.execute(driver_command, params)
  File "/Users/superuser/PycharmProjects/MobileAutoTest/venv/lib/python3.6/site-packages/selenium/webdriver/remote/remote_connection.py", line 374, in execute
    data = utils.dump_json(params)
  File "/Users/superuser/PycharmProjects/MobileAutoTest/venv/lib/python3.6/site-packages/selenium/webdriver/remote/utils.py", line 33, in dump_json
    return json.dumps(json_struct)
[...]

PR #267, opened 2018-10-31
fix python3 set_clipboard error

KazuCocoa:
Python 3 requires encoding in `bytes`. Python 2 doesn't.
close https://github.com/appium/python-client/issues/265

## Python 3

```
>>> driver = webdriver.Remote('http://localhost:4723/wd/hub', desired_caps)
>>> driver.set_clipboard_text("hello")
>>> b = driver.get_clipboard()
>>> b.decode()
'hello'
>>> driver.set_clipboard_text("ありがとう")
>>> b = driver.get_clipboard()
>>> b.decode()
'ありがとう'
```

## Python 2

```
>>> driver = webdriver.Remote('http://localhost:4723/wd/hub', desired_caps)
>>> driver.set_clipboard_text("ありがとう")
>>> b = driver.get_clipboard()
>>> print(unicode(b.decode('UTF-8')))
ありがとう
>>> driver.set_clipboard_text("hello")
>>> b = driver.get_clipboard()
>>> print(unicode(b.decode('UTF-8')))
hello
```

### DIFF
--- a/appium/webdriver/webdriver.py
+++ b/appium/webdriver/webdriver.py
@@ -1288,9 +1288,9 @@ class WebDriver(webdriver.Remote):
 
         def _bytes(value, encoding):
             try:
-                return bytes(value, encoding) # Python 3
+                return bytes(value, encoding)  # Python 3
             except TypeError:
-                return value # Python 2
+                return value  # Python 2
 
         self.set_clipboard(_bytes(str(text), 'UTF-8'), ClipboardContentType.PLAINTEXT, label)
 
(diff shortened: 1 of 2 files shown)


appium/webdriver/webdriver.py
```diff
@@ -1285,7 +1285,14 @@ def set_clipboard_text(self, text, label=None):
         :param text: The text to be set
         :param label: Optional label argument, which only works for Android
         """
-        self.set_clipboard(bytes(text.encode('UTF-8')), ClipboardContentType.PLAINTEXT, label)
+
+        def _bytes(value, encoding):
```
mykola-mokhnach: does it make sense to put this helper method into some separate helper module?

KazuCocoa: I thought so, but nothing method used `bytes` so far. I'd like to extract here in next time.

Issue #269, opened 2018-11-06
Feature Request: driver.push_file(destination_path, source_path)


Jonahss:
### The problem

Right now, in order to put files on a device, the `driver.push_file(path, base64)` function is used.
It can be a little confusing or complex for people to always read a file from their filesystem, encode as base64, then pass it into this function.

### The Solution

How about creating another way to use this function: `driver.push_file(destination_path, source_path)`? This would be simpler for most users.

This work would be a perfect candidate for a Pull Request from someone who wants to make their first contribution to the Appium Python community :D

(suggested by @jlipps)

Jonahss:
Thanks for taking this on @JavonDavis !

JavonDavis:
No problem! @Jonahss

KazuCocoa:
Published https://pypi.org/project/Appium-Python-Client/0.31/
Thanks @JavonDavis !

JavonDavis:
Awesome! Glad I could contribute 😬

PR #270, opened 2018-11-16
driver.push_file(destination_path, source_path) feature

JavonDavis:
This PR aims to simplify pushing and retrieving files from the simulator and emulator. It provides an optional parameter source_path to the existing push_file function and does the base64 conversion on that file for users. 

Fixes https://github.com/appium/python-client/issues/269

To test: 
/Users/javon/Projects/python-client/test/functional/android/push_file_tests.py
/Users/javon/Projects/python-client/test/functional/ios/push_file_tests.py


mykola-mokhnach:
Did you test the PR for both Python 2 and Python3?
I remember there was an issue while base-64 encoding a string there. See https://github.com/appium/python-client/pull/267 for more details.

appium/webdriver/webdriver.py
```diff
@@ -799,15 +795,24 @@ def pull_folder(self, path):
...
-         - path - the path on the device
+         - destination_path - the path on the device
          - base64data - data, encoded as Base64, to be written to the file
+         - source_path - path to file to be loaded on device
         """
+        if source_path is None and base64data is None:
+            raise InvalidArgumentException('Must either present base64 data or a source path on local system')
+
+        if source_path is not None:
+            data = open(source_path, 'rb').read()
```
mykola-mokhnach: I'd rather prefer using `with` for reading the file

appium/webdriver/webdriver.py
```diff
@@ -799,15 +795,24 @@ def pull_folder(self, path):
...
 
-    def push_file(self, path, base64data):
-        """Puts the data, encoded as Base64, in the file specified as `path`.
+    def push_file(self, destination_path, base64data=None, source_path=None):
+        """Puts the data from the file at `source_path`, encoded as Base64, in the file specified as `path`.
 
+        Specify either `base64data` or `source_path`, if both specified default to `source_path`
         :Args:
-         - path - the path on the device
+         - destination_path - the path on the device
```
mykola-mokhnach: I'd rather add more advanced docstrinmg here, since several formatting options are available there. Check https://github.com/appium/java-client/blob/f527be18735f8a60b4f6244f03a1b9954c23a60f/src/main/java/io/appium/java_client/InteractsWithFiles.java#L44

and 

https://github.com/appium/appium-xcuitest-driver/blob/31059c4169516098d2b64a328197a53000ecd360/lib/commands/file-movement.js#L186

for more details

appium/webdriver/webdriver.py
```diff
@@ -799,15 +795,24 @@ def pull_folder(self, path):
...
          - base64data - data, encoded as Base64, to be written to the file
+         - source_path - path to file to be loaded on device
         """
+        if source_path is None and base64data is None:
+            raise InvalidArgumentException('Must either present base64 data or a source path on local system')
+
+        if source_path is not None:
```
mykola-mokhnach: should we check if the file exists first?

appium/webdriver/webdriver.py
jlipps: agree that a context is a better approach so we make sure to clean up file handlers.

appium/webdriver/webdriver.py
jlipps: i think we should, either that or catch the exception and provide something meaningful as a user facing error

test/functional/android/push_file_tests.py
```diff
@@ -0,0 +1,51 @@
...
+
+from appium import webdriver
+
+
+class PushFileTests(unittest.TestCase):
+    def setUp(self):
+        desired_caps = desired_capabilities.get_desired_capabilities('ApiDemos-debug.apk')
+        self.driver = webdriver.Remote('http://localhost:4723/wd/hub', desired_caps)
+
+        # relax template matching
```
jlipps: it looks like the next few lines are unrelated to this test; maybe they were copy and pasted from a different example?

KazuCocoa: Yes, we can remove below settings since they only affect image related features.

appium/webdriver/webdriver.py
mvgiacomello: You could use `if bool(source_path):`

test/functional/android/desired_capabilities.py
```diff
@@ -27,7 +27,8 @@ def get_desired_capabilities(app):
         'platformName': 'Android',
         'deviceName': 'Android Emulator',
         'app': PATH('../../apps/' + app),
-        'newCommandTimeout': 240
+        'newCommandTimeout': 240,
+        'automationName': 'UIAutomator2'
     }
 
     return desired_caps
```
mvgiacomello: Do we use newline at end of file as per PEP8?

appium/webdriver/webdriver.py
JavonDavis: You're right, I've gone ahead and made the fix

appium/webdriver/webdriver.py
JavonDavis: Sure I update the formatting to the reStructured text format, let me know if that one's good. Originally I was just going along with the format I saw in the rest of the file to stay consistent but now I'm noticing some other functions use this format as well.

appium/webdriver/webdriver.py
JavonDavis: The error that it threw was in this is `FileNotFoundError: [Errno 2] No such file or directory: '{filename}', I'm not sure the message I'm presenting now is much different `source_path {} could not be found. Are you sure the file exists?` Is there a better way to do this or this should be good enough?

test/functional/android/push_file_tests.py
JavonDavis: My bad here, you're right @jlipps this is leftovers from a different example

appium/webdriver/webdriver.py
JavonDavis: Not sure there's a real benefit to converting to a boolean for the check?

test/functional/android/desired_capabilities.py
JavonDavis: It passes the pep8 check locally and I didn't make any change at the end of this file just really the addition of UIAutomator2

JavonDavis:
> Did you test the PR for both Python 2 and Python3?
> I remember there was an issue while base-64 encoding a string there. See #267 for more details.

Hey @mykola-mokhnach thanks for the review and I actually ran into that issue while I was working on this, encoding with `base64.b64encode(data).decode('utf-8')` seemed to work well across both the Python versions I tested with (2.7 and 3.7)

test/functional/android/desired_capabilities.py
KazuCocoa: `python -m autopep8 -r --global-config .config-pep8 -d .` runs on CI. I've relaxed the check for `max-line-length`, only. Is it enough for pep8?
(If I missed something, I'd like to add the check on CI)

JavonDavis: @KazuCocoa I think the relaxed 120 max line length is pretty good

appium/webdriver/webdriver.py
```diff
@@ -776,16 +773,15 @@ def set_value(self, element, value):
...
+        """Retrieves the file at `path`. Returns the file's contents.
 
         :Args:
          - path - the path to the file on the device
         """
         data = {
             'path': path,
         }
-        return self.execute(Command.PULL_FILE, data)['value']
+        return base64.b64decode(self.execute(Command.PULL_FILE, data)['value'])
```
mykola-mokhnach: I would say we should not do this, since it is going to break the backward compatibility.

appium/webdriver/webdriver.py
```diff
@@ -799,15 +795,29 @@ def pull_folder(self, path):
...
-    def push_file(self, path, base64data):
-        """Puts the data, encoded as Base64, in the file specified as `path`.
+    def push_file(self, destination_path, base64data=None, source_path=None):
+        """Puts the data from the file at `source_path`, encoded as Base64, in the file specified as `path`.
 
-        :Args:
-         - path - the path on the device
-         - base64data - data, encoded as Base64, to be written to the file
+        Specify either `base64data` or `source_path`, if both specified default to `source_path`
+        :param destination_path: the path on the device/simulator
```
mykola-mokhnach: Please add more detailed description of this argument

appium/webdriver/webdriver.py
JavonDavis: You're right I overlooked this, it's removed

appium/webdriver/webdriver.py
JavonDavis: Is the new description a little better?

appium/webdriver/webdriver.py
jlipps: I like the current state of the code (with the slightly more informative error message). Thanks.

KazuCocoa:
Thanks! Will merge this.

PR #271, opened 2018-11-22
append document for recording screen

KazuCocoa:
Python does not need to add parameters since options send to the server directly

### DIFF
--- a/appium/webdriver/webdriver.py
+++ b/appium/webdriver/webdriver.py
@@ -1220,7 +1220,6 @@ class WebDriver(webdriver.Remote):
 
         iOS Specific:
         - videoQuality: The video encoding quality: 'low', 'medium', 'high', 'photo'. Defaults to 'medium'.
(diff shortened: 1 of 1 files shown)


appium/webdriver/webdriver.py
```diff
@@ -1222,8 +1222,12 @@ def start_recording_screen(self, **options):
         - videoQuality: The video encoding quality: 'low', 'medium', 'high', 'photo'. Defaults to 'medium'.
         Only works for real devices.
         - videoType: The format of the screen capture to be recorded.
-        Available formats: 'h264', 'mp4' or 'fmp4'. Default is 'mp4'.
-        Only works for Simulator.
+        Available formats: Execute `ffmpeg -codecs` in the terminal to see the list of supported video codecs.
```
mykola-mokhnach: It makes sense to mention these values only work since Appium 1.10.0

PR #274, opened 2018-11-24
updated dependencies in requirements.txt file with version number

RajeshkumarAyyadurai:
As a best practice, updated dependencies in requirements.txt file with version for each library


RajeshkumarAyyadurai:
pylint 2.1.1 is also not supported to python 2.7. I will again update all the libraries in requirements.txt file with python 2.7 supported.

PR #278, opened 2018-11-25
Fixed few failing tests in appium_tests.py

RajeshkumarAyyadurai:
There are 12 test cases failed in appium_tests.py. Fixed 4 test cases for now, that are listed below:

1. test_background_app
2. test_close__and_launch_app
3. test_reset
4. test_send_keys

### DIFF
--- a/test/functional/android/appium_tests.py
+++ b/test/functional/android/appium_tests.py
@@ -131,7 +131,7 @@ class AppiumTests(unittest.TestCase):
     def test_background_app(self):
         self.driver.background_app(1)
         sleep(5)
-        el = self.driver.find_element_by_name('Animation')
+        el = self.driver.find_element_by_android_uiautomator('new UiSelector().text("Animation")')
         self.assertIsNotNone(el)
 
     def test_is_app_installed(self):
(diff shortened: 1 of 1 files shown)


test/functional/android/appium_tests.py
```diff
@@ -131,7 +131,7 @@ def test_pull_folder(self):
     def test_background_app(self):
         self.driver.background_app(1)
         sleep(5)
-        el = self.driver.find_element_by_name('Animation')
```
mykola-mokhnach: name locators have been replaced with accessibility id

test/functional/android/appium_tests.py
```diff
@@ -131,7 +131,7 @@ def test_pull_folder(self):
...
+        el = self.driver.find_element_by_android_uiautomator('new UiSelector().text("Animation")')
```
KazuCocoa: I'd rather use non-uiautomator locator as possible since the locator does not work on Espresso driver.
Can we use another `find_element_by_x`?

RajeshkumarAyyadurai: Yeah, I understand. Updated the tests again with non-uiautomator locator strategy. Thanks for the info!

test/functional/android/appium_tests.py
RajeshkumarAyyadurai: Ok, got it. Thanks.

test/functional/android/appium_tests.py
KazuCocoa: thanks :)

PR #280, opened 2018-11-27
Update setup elements

KazuCocoa:
1. Removed `docgen.py` and `readme_gen.py`
    - They were introduced in https://github.com/appium/python-client/commit/ae0dd8e7983cdb89af484cb551bc918afdc09cb4#diff-6276ffbde68fe0e9c7c563ee33ab8ee7 . I guess we would have liked to use them as long description in pypi since they only converted markdwon formatted README into another format
    - But Pipy has already supported markdown format. So, we probably can use current `README.md` as the long description without any modifications.
2. Update `classifiers` following selenium-python binding.


setup.py
```diff
@@ -41,8 +43,12 @@
     ],
     license='Apache 2.0',
     classifiers=[
-        'Development Status :: 3 - Alpha',
+        'Development Status :: 4 - Beta',
         'Programming Language :: Python',
+        'Programming Language :: Python :: 2.7',
+        'Programming Language :: Python :: 3.4',
+        'Programming Language :: Python :: 3.5',
+        'Programming Language :: Python :: 3.6',
```
mykola-mokhnach: 3.7?

KazuCocoa:
@imurchie Can you take a look when you have a time?
I think you can remember the purpose of `docgen.py` and `readme_gen.py`.

PR #281, opened 2018-11-28
Introduce httpretty for unittest to mock Appium server

KazuCocoa:
As Python, we must take care of compatibility of 2.x and 3.x.
e.g. https://github.com/appium/python-client/pull/267/files

Currently, this project has many functional tests which need Appium server. But we can uncover such issues with dynamic tests like this PR's approach, `test_clipboard`. This approach is using a mock server to simulate Appium server using _httpretty_ for unit tests. We do not need to run the functional tests every PR to ensure the compatibility of Python 2 and 3 with this approach.

If we have no https://github.com/appium/python-client/pull/267/files , the test case can fail like below with Python 3.x 

```
======================================================================
ERROR: test_clipboard (__main__.WebDriverWebDriverTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/httpretty/core.py", line 1702, in wrapper
    return test(*args, **kw)
  File "test/unit/webdriver/webdriver_tests.py", line 77, in test_clipboard
    driver.set_clipboard_text('hello')
  File "/usr/local/lib/python3.7/site-packages/appium/webdriver/webdriver.py", line 1305, in set_clipboard_text
    self.set_clipboard(str(text), ClipboardContentType.PLAINTEXT, label)
  File "/usr/local/lib/python3.7/site-packages/appium/webdriver/webdriver.py", line 1284, in set_clipboard
    'content': base64.b64encode(content).decode('UTF-8'),
  File "/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/lib/python3.7/base64.py", line 58, in b64encode
    encoded = binascii.b2a_base64(s, newline=False)
TypeError: a bytes-like object is required, not 'str'
```

I have maintained this approach in ruby_lib_core. It works fine to ensure client-side behaviour stably and short.

I would like to clean this code if this approach is reasonable for us.

----

I will ask this approach in slack channel.


jlipps:
This is a good idea. You could also write tests against Appium's FakeDriver to get a "real" driver that does nothing :-)

KazuCocoa:
ah, gotha. => https://github.com/appium/appium/blob/master/lib/appium.js#L30
I inputted the soce in my brain 👍

Issue #283, opened 2018-12-01
Add unit tests


KazuCocoa:
Currently, almost test cases for this repository depend on functional tests which require Appium server. But as a client, we can make sure each method and their compatibility of Python 2 and 3 in unit test level.

Example: https://github.com/appium/python-client/pull/281 , https://github.com/appium/python-client/pull/282

Adding unit tests will be a good entry point to commit Appium project as well.
I'd set `good first issue` label.

KazuCocoa:
Below is added in this library

- [ ] ('getContexts',)
- [ ] ('getCurrentContext',)
- [ ] touchAction
- [ ] multiAction
- [ ] getAppStrings
- [ ] keyEvent
- [ ] pressKeyCode
- [ ] longPressKeyCode
- [ ] getCurrentActivity
- [ ] getCurrentPackage
- [ ] setImmediateValue
- [ ] pullFile
- [ ] pullFolder
- [ ] pushFile
- [ ] background
- [ ] isAppInstalled
- [ ] installApp
- [ ] removeApp
- [ ] terminateApp
- [ ] activateApp
- [ ] queryAppState
- [ ] startActivity
- [ ] launchApp
- [ ] closeApp
- [ ] endTestCoverage
- [ ] lock
- [ ] unlock
- [ ] isLocked
- [ ] shake
- [ ] touchId
- [ ] toggleTouchIdEnrollment
- [ ] reset
- [ ] hideKeyboard
- [ ] isKeyboardShown
- [ ] openNotifications
- [ ] getAvailableIMEEngines
- [ ] isIMEActive
- [ ] activateIMEEngine
- [ ] deactivateIMEEngine
- [ ] getActiveEngine
- [ ] replaceKeys
- [ ] getSettings
- [ ] updateSettings
- [ ] toggleLocationServices
- [ ] toggleWiFi
- [ ] locationInView
- [ ] getDeviceTime
- [ ] clear
- [ ] startRecordingScreen
- [ ] stopRecordingScreen
- [ ] setClipboard
- [ ] getClipboard
- [ ] compareImages
- [ ] fingerPrint

PR #284, opened 2018-12-04
Improve pytest

KazuCocoa:
A part of https://github.com/appium/python-client/issues/283 

- Add `test/pytest.ini` to make the result simple
- Add `ios_w3c_driver` to get a driver for iOS
- Define base url and generating command URL method


test/unit/webdriver/webdriver_test.py
```diff
@@ -82,3 +82,17 @@ def test_create_session_forceMjsonwp(self):
 
         assert driver.session_id == 'session-id'
         assert driver.w3c is False
+
+    @httpretty.activate
+    def test_a_number_of_commands_w3c_android(self):
+        """Track how many commands defined for Android W3C"""
+        driver = android_w3c_driver()
+        # pylint: disable=W0212
+        assert len(driver.command_executor._commands.keys()) == 174
```
mykola-mokhnach: I would say it is an extra effort to always update this number when a new command is added/removed

KazuCocoa: Definitely, I thought so. Will remove it

PR #288, opened 2018-12-13
Add unit tests for isLocked Library

VenkateshPS:
Add unit tests for isLocked Library


test/unit/webdriver/device/islocked_test.py
```diff
@@ -0,0 +1,41 @@
...
+
+    @httpretty.activate
+    def test_islocked_false(self):
+        driver = android_w3c_driver()
+        httpretty.register_uri(
+            httpretty.POST,
+            appium_command('/session/1234567890/appium/device/is_locked'),
+            body='{"value": false}'
+        )
+        assert driver.is_locked() == False
```
KazuCocoa: What about `assert xxxx is False` syntax

VenkateshPS: What about assert xxxx is False syntax ?

KazuCocoa: https://stackoverflow.com/questions/27276610/boolean-identity-true-vs-is-true
`is`, `is not`

VenkateshPS: Thanks done

Issue #289, opened 2018-12-16
AttributeError: 'WebDriver' object has no attribute 'location'


Gilad-WT:
## The problem

I tried using the driver.location() function and got AttributeError: 'WebDriver' object has no attribute 'get_location'. I then checked the set_location() and that seems to work so I looked at the code in /appium/webdriver/webdriver.py and found that the function doesn't exist anywhere in the code. I think it was probably accidentally deleted at some point. Can you please confirm this is the case?

Also while checking the set_location() function I found that no mater what altitude is passed to the function it sets the emulator's altitude to 0.0.  Is altitude changing supported? I think this is a different bug but they are both related to Geo location so I put these issues together.

## Environment
- Appium version 1.82
- Desktop OS/version used to run Appium: Windows 10
- Mobile platform/version under test: android 7.0
- Real device or emulator/simulator: android emulator api 24
- Appium CLI or Appium.app|exe: desktop version

KazuCocoa:
Can you attach full server side log when you call `set_location`?

Python client has no `get_location` so far. I could find only `set_location` implementation in log.
We need to implement get_location for this client.

KazuCocoa:
https://github.com/appium/python-client/blob/master/CHANGELOG.rst#v032-2018-12-18 includes the `get_location`

KazuCocoa:
The get location feature should work as `#location` property.

PR #291, opened 2018-12-17
Split driver methods into mixin classes

mykola-mokhnach:
I think the main WebDriver class is too large and thus is hard to maintain. It would be smarter to split command groups into separate mixin classes, like, for example, Robot Framework [does](https://github.com/robotframework/robotframework/blob/master/src/robot/libraries/BuiltIn.py)


appium/webdriver/mobilecommand.py
```diff
@@ -66,6 +66,7 @@ class MobileCommand(object):
     GET_SETTINGS = 'getSettings'
     UPDATE_SETTINGS = 'updateSettings'
     SET_LOCATION = 'setLocation'
+    GET_LOCATION = 'getLocation'
```
KazuCocoa: 👍 https://github.com/appium/python-client/issues/289

PR #297, opened 2018-12-18
Fix overridden mixin method call

mykola-mokhnach:
Addresses https://travis-ci.org/appium/python-client/jobs/469424297


appium/webdriver/webdriver.py
```diff
@@ -1412,19 +1412,10 @@ def finger_print(self, finger_id):
...
-            ('POST', '/session/$sessionId/context')
-
-        self.command_executor._commands[Command.TOGGLE_LOCATION_SERVICES] = \
-            ('POST', '/session/$sessionId/appium/device/toggle_location_services')
-        self.command_executor._commands[Command.GET_LOCATION] = \
-            ('GET', '/session/$sessionId/location')
-        self.command_executor._commands[Command.SET_LOCATION] = \
-            ('POST', '/session/$sessionId/location')
+        # call the overridden command binders from all mixin classes
+        for mixin_class in filter(lambda x: x is not self.__class__, self.__class__.__mro__):
```
KazuCocoa: ah... interesting..

according https://docs.quantifiedcode.com/python-anti-patterns/readability/using_map_or_filter_where_list_comprehension_is_possible.html , we also can implement as:
`for mixin_class in [x for x in self.__class__.__mro__ if x is not self.__class__]:`

(Failing autopep8 was probably different point though)

PR #306, opened 2019-01-04
Add touch action unittest

tadashi0713:

### DIFF
--- a/test/unit/webdriver/touch_action_test.py
+++ b/test/unit/webdriver/touch_action_test.py
@@ -88,14 +88,14 @@ class TestTouchAction(object):
 
     def test_release_json(self, touch_action):
         json = [
-            {'action': 'release', 'options': {}}
+            {'action': 'tap', 'options': {'count': 1, 'element': 8}}, {'action': 'release', 'options': {}}
         ]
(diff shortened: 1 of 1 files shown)


test/unit/webdriver/touch_action_test.py
```diff
@@ -37,6 +37,67 @@ def test_tap_x_y_json(self, touch_action):
...
+        json = [
+            {'action': 'tap', 'options': {'count': 1, 'element': 8}}, {'action': 'release', 'options': {}}
+        ]
+        touch_action.tap(ElementStub(8)).release()
+        assert json == touch_action.json_wire_gestures
+
+    def test_perform_json(self, touch_action):
+        json = []
+        touch_action.tap(ElementStub(9)).perform()
+        assert json == touch_action.json_wire_gestures
```
KazuCocoa: You can write as below if you would like to make sure the `json` is exactly cleared after the `perform`

```
json_tap = [...]
touch_action.tap(ElementStub(9))
assert json_tap == touch_action.json_wire_gestures
touch_action.perform()
assert [] == touch_action.json_wire_gestures
```

PR #308, opened 2019-01-05
Add network unittest

ki4070ma:

### DIFF
--- a/test/unit/webdriver/device/network_test.py
+++ b/test/unit/webdriver/device/network_test.py
@@ -16,6 +16,7 @@ from appium.webdriver.webdriver import WebDriver
 from test.unit.helper.test_helper import appium_command, android_w3c_driver
 
 import httpretty
+import json
 
 
 class TestWebDriverNetwork(object):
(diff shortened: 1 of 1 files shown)


test/unit/webdriver/device/network_test.py
```diff
@@ -0,0 +1,50 @@
...
+
+    @httpretty.activate
+    def test_set_network_connection(self):
+        driver = android_w3c_driver()
+        httpretty.register_uri(
+            httpretty.POST,
+            appium_command('/session/1234567890/network_connection'),
+            body='{"value": 2}'
+        )
+        assert driver.set_network_connection(2) == 2
```
KazuCocoa: Could you add assertions for the request body like below?
https://github.com/appium/python-client/blob/master/test/unit/webdriver/device/clipboard_test.py
The return value is enough to `""`

ki4070ma: Thanks for your review.

Done by https://github.com/appium/python-client/pull/308/commits/3d4487ddb03beb97da5d3140d85699bf79547844

PR #309, opened 2019-01-06
Add devicetime unittest

ki4070ma:

### DIFF
--- a/test/unit/webdriver/device/device_time_test.py
+++ b/test/unit/webdriver/device/device_time_test.py
@@ -1,0 +1,31 @@
+#!/usr/bin/env python
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


KazuCocoa:
Thanks. The method can set `format` argument like below.
https://github.com/appium/ruby_lib_core/blob/680577d83356fc19d5fc435d5e49f0fea3c1442a/lib/appium_lib_core/common/base/driver.rb#L613-L627

Can you extend current method format to it if you have a chance?

ki4070ma:
Thanks for your review.
And ok, will do.

test/unit/webdriver/device/device_time_test.py
```diff
@@ -0,0 +1,30 @@
...
+
+    @httpretty.activate
+    def test_device_time(self):
+        driver = android_w3c_driver()
+        httpretty.register_uri(
+            httpretty.GET,
+            appium_command('/session/1234567890/appium/device/system_time'),
+            body='{"value": ""}'
+        )
+        driver.device_time  # Expected no exception
```
KazuCocoa: Can you add one assertion?

ki4070ma: Thanks, my assertion was not enough. 
I've changed here to check if return value is same to value in body.

Issue #311, opened 2019-01-07
Can't set format to device_time as argument


ki4070ma:
Needed following 2 changes

1. Add ```format``` to device_time
   * https://github.com/appium/python-client/blob/master/appium/webdriver/webdriver.py
1. Enabled 'GET' to have body
   * https://github.com/SeleniumHQ/selenium/blob/a920faac5532175843c6a135f1f2dd621d8016c3/py/selenium/webdriver/remote/remote_connection.py#L393

For now, I'm not sure reason for No.2.
Need further checks.

KazuCocoa:
ah.. I see.

It is okay to suspend the change.
I will change appium server side following http format.
(Accept POST, for example)

ki4070ma:
Thanks to https://github.com/appium/appium-base-driver/pull/294, considering No.2 is unnecessaey.

PR #307, opened 2019-01-04
Move action and keyboard helpers to mixin classes

mykola-mokhnach:

PR #312, opened 2019-01-08
Added format to device_time as argument

ki4070ma:
For https://github.com/appium/python-client/issues/311

Which option should I take? (Maybe No.2?)

1. Removed property decorator from device_time
   * https://github.com/appium/python-client/commit/c2c5f9889bb128b147c43065f9fab68766236cd9
   * [Pros] Normal api style
   * [Cons] Impact to current users
1. Added another api like device_time_format()
   * [Pros] No impact to current users
   * [Cons] Increased similar api (e.g. device_time, device_time_format())


mykola-mokhnach:
How about the third option - we keep the property and add a new setter for datetime format?

ki4070ma:
Thanks for your comment.
Do you mean that 'new setter' is for property?
If yes, setter for property can't be used in my understanding.

https://docs.python.org/3/library/functions.html?highlight=property#property

if no, could you tell more for details?

appium/webdriver/webdriver.py
```diff
@@ -585,6 +603,6 @@ def _addCommands(self):
         self.command_executor._commands[Command.LOCATION_IN_VIEW] = \
             ('GET', '/session/$sessionId/element/$id/location_in_view')
         self.command_executor._commands[Command.GET_DEVICE_TIME] = \
-            ('GET', '/session/$sessionId/appium/device/system_time')
```
ki4070ma: @KazuCocoa 

Can I change to only 'POST'?
My motivation is just to avoid complicated naming around below codes.

https://github.com/appium/python-client/blob/a40a933a906fe4cc8d31c83d55ba52f4830242b3/appium/webdriver/webdriver.py#L587-L588

e.g.
```
 self.command_executor._commands[Command.GET_DEVICE_TIME_GET] = \ 
('GET', '/session/$sessionId/appium/device/system_time') 
 self.command_executor._commands[Command.GET_DEVICE_TIME_POST] = \ 
('POST', '/session/$sessionId/appium/device/system_time') 
```
Your idea will be different from other variable naming..

If GET behavior should be kept, I'll update so.
(I'm not sure the impact to current behavior  by change from POST to GET.)

Thanks.

KazuCocoa: You should define another command like your _e.g._ as compatibility for released Appium servers.
Current released appium server cannot handle the post method. We observe users who use older Appium servers. Other libraries also use this library in their libs.

Thus, we would like to avoid breaking changes as possible.
(Instead, please leave a comment we can remove it in the future, for instance.)

mykola-mokhnach: I'd also keep the old GET endpoint. I was thinking on that and I think it is more logical to leave the existing property as as and add `get_formatted_device_time(self, format)` method, which is going to call the new POST. Also, in the documentation for this method it is necessary to mention it is only available since Appium 1.11

ki4070ma: @KazuCocoa @mykola-mokhnach 
Thanks a lot, you're right. I should have considered more.
I'll update along to your comments.

appium/webdriver/webdriver.py
```diff
@@ -523,9 +523,23 @@ def update_settings(self, settings):
...
     @property
     def device_time(self):
-        """Returns the date and time from the device
+        """Returns the date and time from the device.
         """
-        return self.execute(Command.GET_DEVICE_TIME, {})['value']
+        return self.execute(Command.GET_DEVICE_TIME_GET, {})['value']
+
+    def get_device_time(self, format=''):
+        """Returns the date and time from the device. (Only available since Appium 1.11.0)
```
ki4070ma: TODO: Need to check next released Appium server version which includes https://github.com/appium/appium-base-driver/pull/294

appium/webdriver/webdriver.py
```diff
@@ -523,9 +523,23 @@ def update_settings(self, settings):
 
     @property
     def device_time(self):
-        """Returns the date and time from the device
+        """Returns the date and time from the device.
         """
-        return self.execute(Command.GET_DEVICE_TIME, {})['value']
+        return self.execute(Command.GET_DEVICE_TIME_GET, {})['value']
+
+    def get_device_time(self, format=''):
```
mykola-mokhnach: the default value should be None

appium/webdriver/webdriver.py
```diff
@@ -523,9 +523,23 @@ def update_settings(self, settings):
...
+
+        :Args:
+         - format - The default format is `YYYY-MM-DDTHH:mm:ssZ`, which complies to ISO-8601
+           The set of format specifiers. Read https://momentjs.com/docs/ to get
+           the full list of supported datetime format specifiers.
+
+        :Usage:
+            self.driver.get_device_time(self, "YYYY-MM-DD")
+        """
+        data = {'format': format} if format else {}
```
mykola-mokhnach: if format is None then device_time should be returned

appium/webdriver/webdriver.py
ki4070ma: Thanks. Yes, I've checked behavior with my changes in my local.
My TODO means I'll need to update version on comment if next version is different from 1.11.0. (e.g. 1.10.1)

If my understanding is wrong, please let me know.

Thanks.

ki4070ma: Thanks, I've removed 'DO NOT MERGE' from issue's title.

appium/webdriver/webdriver.py
```diff
@@ -523,9 +523,25 @@ def update_settings(self, settings):
...
+         - format - The default format is `YYYY-MM-DDTHH:mm:ssZ`, which complies to ISO-8601
+           The set of format specifiers. Read https://momentjs.com/docs/ to get
+           the full list of supported datetime format specifiers.
+
+        :Usage:
+            self.driver.get_device_time("YYYY-MM-DD")
+        """
+        if format:
+            return self.execute(Command.GET_DEVICE_TIME_POST, {'format': format})['value']
+        else:
```
mykola-mokhnach: if format is None (https://docs.quantifiedcode.com/python-anti-patterns/readability/comparison_to_none.html)

also, `else` is redundant

appium/webdriver/webdriver.py
```diff
@@ -523,9 +523,25 @@ def update_settings(self, settings):
...
-        return self.execute(Command.GET_DEVICE_TIME, {})['value']
+        return self.execute(Command.GET_DEVICE_TIME_GET, {})['value']
+
+    def get_device_time(self, format=None):
+        """Returns the date and time from the device. (Only available since Appium 1.11.0)
+
+        :Args:
+         - format - The default format is `YYYY-MM-DDTHH:mm:ssZ`, which complies to ISO-8601
+           The set of format specifiers. Read https://momentjs.com/docs/ to get
+           the full list of supported datetime format specifiers.
```
mykola-mokhnach: please also add that if unset then the result of `device_time` is returned by default

appium/webdriver/webdriver.py
ki4070ma:
```
if format is None:
    return self.device_time
return self.execute(Command.GET_DEVICE_TIME_POST, {'format': format})['value']
```
is better than below?
```
if format:
    return  self.execute(Command.GET_DEVICE_TIME_POST, {'format': format})['value']
return self.device_time
```

>also, else is redundant

oh,,,thanks.

appium/webdriver/webdriver.py
ki4070ma: Done by https://github.com/appium/python-client/pull/312/commits/771324b0d9d36a4da3d7ee8bce3a4c0999f6d58d

appium/webdriver/webdriver.py
ki4070ma: Done by https://github.com/appium/python-client/pull/312/commits/54815e8dd04947a2738ddc98ecb3e7cecdebec02

KazuCocoa: The PR indicates the default argument is None. It is good to handle return value following the None then. None is better than `""`, since we can compare them as literal explicitly.

mykola-mokhnach:
I would also move the device-time related methods to a separate extension class, like in 
https://github.com/appium/python-client/pull/307

mykola-mokhnach:
Although, this can be done in the other PR

PR #322, opened 2019-01-31
add pressure option

KazuCocoa:
Tap by coordinates doesn't work with iOS devices using 3DTouch #321 

Will add test too

### DIFF
--- a/appium/webdriver/common/touch_action.py
+++ b/appium/webdriver/common/touch_action.py
@@ -49,10 +49,10 @@ class TouchAction(object):
 
         return self
 
-    def press(self, el=None, x=None, y=None):
+    def press(self, el=None, x=None, y=None, pressure=None):
         """Begin a chain with a press down action at a particular element or point
         """
-        self._add_action('press', self._get_opts(el, x, y))
+        self._add_action('press', self._get_opts_with_pressure(el, x, y, pressure=pressure))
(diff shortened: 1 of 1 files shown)


appium/webdriver/common/touch_action.py
```diff
@@ -49,10 +49,10 @@ def tap(self, element=None, x=None, y=None, count=1):
 
         return self
 
-    def press(self, el=None, x=None, y=None):
+    def press(self, el=None, x=None, y=None, pressure=None):
```
mykola-mokhnach: It would be nice to describe what the pressure value is

appium/webdriver/common/touch_action.py
```diff
@@ -49,10 +49,17 @@ def tap(self, element=None, x=None, y=None, count=1):
...
         """Begin a chain with a press down action at a particular element or point
+
+        :Args:
+        - el - (optional) the element to press
+        - x - (optional) x coordiate to press. If y is used, x must also be set
+        - y - (optional) y coordiate to press. If x is used, y must also be set
+        - pressure - (optional) [iOS Only] press as force touch. Read the description of `force` property on Apple's UITouch cla
```
mykola-mokhnach: what is `Apple's UITouch cla` ?

PR #338, opened 2019-02-25
Add direct connect flag to be able to handle directConnectXxxxc

KazuCocoa:
This feature is the same as https://github.com/projectxyzio/web2driver#direct-connect-urls and https://www.rubydoc.info/github/appium/ruby_lib_core/Appium/Core/Driver#direct_connect-instance_method

Use case:

- If a user has a proxy server such as SeleniumGrid as below, an Appium client communicate with Appium servers via the proxy every command.
    ```
    # via proxy server in session creation
    client <--> proxy server <---> appium server <-> devices
                                                 `-> appium server <-> devices
    ```
    - Then, the communication takes additional seconds between the client and servers. The proxy server is if in the US and the Appium server is in Europ, the time increases more.
    - To reduce the delay, we can distribute the proxy server to each reason, but it increases the complexity of the proxy server management.
- If we can make such communication directly with the client and servers, it can reduce the time easily.
    ```
    # communicate directly after the session creation
    client <--------> appium server <-> devices
                       `---> appium server <-> devices
    ```
    - The first create session command takes the additional time, but the following commands can communicate with servers directly with less communication latency.
        - We can manage a bunch of custom urls in a client side, but it increases management complexity in the client side.

I've added `direct_connection` flag, and it is false by default, to be able to handle the flag.

Different as SeleniumGrid, mobile apps tie to physical locations. Measuring performance including network latency is necessary to manage devices in various locations, for example. In such case, one proxy server and Appium servers in various locations make easy to manage the complexity. The proxy has device selection logic. The client only sends a session creation command to the proxy, no many URLs.

---

cc @jlipps

### DIFF
--- a/appium/webdriver/webdriver.py
+++ b/appium/webdriver/webdriver.py
@@ -20,6 +20,8 @@ import copy
 from selenium.common.exceptions import InvalidArgumentException
 from selenium.webdriver.common.by import By
 from selenium.webdriver.remote.command import Command as RemoteCommand
+from selenium.webdriver.remote.remote_connection import RemoteConnection
+
 
 from appium.webdriver.common.mobileby import MobileBy
 from .appium_connection import AppiumConnection
@@ -117,7 +119,7 @@ class WebDriver(
 ):
 
     def __init__(self, command_executor='http://127.0.0.1:4444/wd/hub',
-                 desired_capabilities=None, browser_profile=None, proxy=None, keep_alive=False):
+                 desired_capabilities=None, browser_profile=None, proxy=None, keep_alive=False, direct_connection=False):
 
         super(WebDriver, self).__init__(
             AppiumConnection(command_executor, keep_alive=keep_alive),
@@ -126,11 +128,27 @@ class WebDriver(
             proxy
         )
 
-        if self.command_executor is not None:
+        if self.command_executor is not None:  # pylint: disable=access-member-before-definition
             self._addCommands()
 
         self.error_handler = MobileErrorHandler()
         self._switch_to = MobileSwitchTo(self)
+
+        if direct_connection:
+            protocol = self.capabilities['directConnectProtocol']
+            hostname = self.capabilities['directConnectHost']
+            port = self.capabilities['directConnectPort']
+            path = self.capabilities['directConnectPath']
+            # TODO: create a new client to the above params
+            executor = '{scheme}://{hostname}:{port}{path}'.format(
+                scheme=protocol,
+                hostname=hostname,
+                port=port,
+                path=path
+            )
+            # Override command executor
+            self.command_executor = RemoteConnection(executor, keep_alive=keep_alive)
+            self._addCommands()
 
         # add new method to the `find_by_*` pantheon
         By.IOS_UIAUTOMATION = MobileBy.IOS_UIAUTOMATION
--- a/appium/webdriver/webdriver.py
+++ b/appium/webdriver/webdriver.py
@@ -129,7 +129,7 @@ class WebDriver(
             proxy
         )
 
-        if self.command_executor is not None:  # pylint: disable=access-member-before-definition
+        if hasattr(self, 'command_executor'):
             self._addCommands()
 
         self.error_handler = MobileErrorHandler()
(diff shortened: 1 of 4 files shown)


appium/webdriver/webdriver.py
```diff
@@ -142,6 +148,24 @@ def __init__(self, command_executor='http://127.0.0.1:4444/wd/hub',
         By.IMAGE = MobileBy.IMAGE
         By.CUSTOM = MobileBy.CUSTOM
 
+    def _update_command_executor(self, keep_alive):
```
jlipps: should probably check that these values actually exist. just because someone has set the `direct_connection` param doesn't mean the server supports it, right?

up to you whether you think we should throw an error if one of these values is not present, or just silently ignore the request to direct connect.

appium/webdriver/webdriver.py
```diff
@@ -142,6 +148,24 @@ def __init__(self, command_executor='http://127.0.0.1:4444/wd/hub',
...
+        port = self.capabilities['directConnectPort']
+        path = self.capabilities['directConnectPath']
+        executor = '{scheme}://{hostname}:{port}{path}'.format(
+            scheme=protocol,
+            hostname=hostname,
+            port=port,
+            path=path
+        )
+
+        logger.info('Update request endpoint to %s', executor)
```
mykola-mokhnach: Updated

appium/webdriver/webdriver.py
```diff
@@ -142,6 +148,24 @@ def __init__(self, command_executor='http://127.0.0.1:4444/wd/hub',
...
+        # Override command executor
+        self.command_executor = RemoteConnection(executor, keep_alive=keep_alive)
+        self._addCommands()
```
mykola-mokhnach: do we want to execute _addCommands twice?

appium/webdriver/webdriver.py
KazuCocoa: nice catch, I forgot it...

I'll show warning or error message rather than stopping test script since whether a server cannot handle the following requests or not depend on them.

appium/webdriver/webdriver.py
KazuCocoa: Yes.
`self.command_executor = RemoteConnection(executor, keep_alive=keep_alive)` replaces executor with new instance. The new one does not have commands added by `_addCommands()` ( https://github.com/appium/python-client/pull/338/files#diff-d2ce7e5cde58c87021c848f4b40b2106R216 fails if we do not add this line. )

appium/webdriver/webdriver.py
```diff
@@ -126,12 +129,15 @@ def __init__(self, command_executor='http://127.0.0.1:4444/wd/hub',
             proxy
         )
 
-        if self.command_executor is not None:
+        if self.command_executor is not None:  # pylint: disable=access-member-before-definition
```
mykola-mokhnach: you could probably  use hasattr to avoid pylint warning

appium/webdriver/webdriver.py
```diff
@@ -142,6 +148,40 @@ def __init__(self, command_executor='http://127.0.0.1:4444/wd/hub',
...
+        """Update command executor following directConnect feature"""
+        direct_protocol = 'directConnectProtocol'
+        direct_host = 'directConnectHost'
+        direct_port = 'directConnectPort'
+        direct_path = 'directConnectPath'
+
+        if (not {direct_protocol, direct_host, direct_port, direct_path}.issubset(set(self.capabilities))):
+            message = 'Direct connect capabilities from server were:\n'\
+                '{protocol}: {protocol_get},  {host}: {host_get}, {port}: {port_get}, {path}: {path_get}'.format(
+                    protocol=direct_protocol, protocol_get=self.capabilities.get(direct_protocol, ''),
```
mykola-mokhnach: capabilities is already a dictionary, so we could in theory to something like `**self.capabilities`

KazuCocoa: I'd like to show a blank in no key case like below, so I changed to `for` style

```
directConnectProtocol: 'http'
directConnectHost: 'localhost2'
directConnectPort: '4800'
directConnectPath: ''
```

PR #352, opened 2019-04-22
Fix True/False in image settings, add boolean value in settings test

KazuCocoa:
True/False in https://github.com/appium/appium/blob/c9a9aa354e203cc8d8c869bae0947aff328f3f01/docs/en/advanced-concepts/image-elements.md should be boolean

### DIFF
--- a/test/unit/webdriver/device/settings_test.py
+++ b/test/unit/webdriver/device/settings_test.py
@@ -26,23 +26,45 @@ from appium.webdriver.webdriver import WebDriver
 class TestWebDriverSettings(object):
 
     @httpretty.activate
-    def test_get_settings(self):
+    def test_get_settings_bool(self):
         driver = android_w3c_driver()
         httpretty.register_uri(
             httpretty.GET,
             appium_command('/session/1234567890/appium/settings'),
-            body='{"value": {"sample": 123}}'
+            body='{"value": {"sample": true}}'
         )
-        assert driver.get_settings()['sample'] == 123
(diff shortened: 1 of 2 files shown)


test/unit/webdriver/device/settings_test.py
```diff
@@ -26,23 +26,45 @@
...
+    def test_get_settings_bool(self):
         driver = android_w3c_driver()
         httpretty.register_uri(
             httpretty.GET,
             appium_command('/session/1234567890/appium/settings'),
-            body='{"value": {"sample": 123}}'
+            body='{"value": {"sample": true}}'
         )
-        assert driver.get_settings()['sample'] == 123
+        assert driver.get_settings()['sample'] == True
```
mykola-mokhnach: is True

test/unit/webdriver/device/settings_test.py
```diff
@@ -26,23 +26,45 @@
...
         httpretty.register_uri(
             httpretty.POST,
             appium_command('/session/1234567890/appium/settings'),
         )
-        assert isinstance(driver.update_settings({"sample": 123}), WebDriver) is True
+        assert isinstance(driver.update_settings({"sample": True}), WebDriver) is True
 
         d = get_httpretty_request_body(httpretty.last_request())
-        assert d['settings']['sample'] == 123
+        assert d['settings']['sample'] == True
```
mykola-mokhnach: same here

PR #358, opened 2019-04-29
mobile:pinchOpen and mobile:pinchClose no longer implemented in appiu…

Jonahss:
…m drivers


test/functional/android/multi_action_tests.py
```diff
@@ -1,172 +0,0 @@
...
-
-class MultiActionTests(unittest.TestCase):
-    def setUp(self):
-        desired_caps = desired_capabilities.get_desired_capabilities('ApiDemos-debug.apk')
-        self.driver = webdriver.Remote('http://localhost:4723/wd/hub', desired_caps)
-
-    def tearDown(self):
-        self.driver.quit()
-
-    def test_parallel_actions(self):
```
KazuCocoa: Why did you get rid of entire this test, not only `test_driver_pinch` and `test_driver_zoom`?
(Did not work?)

Jonahss: Ah! I looked at the ios multitouch file and assumed the android one was a duplicate. Thanks for catching! I'll fix it.

Jonahss: Fixed. Just deleting the two tests now.

PR #359, opened 2019-04-30
Add arg checks to power emulation

ki4070ma:


mykola-mokhnach:
I think this is not a good approach to add such verification on the client side. It must be done on server side (which is also questionable though)

ki4070ma:
Thanks for your quick comments. I understand your points.

* clients just send request with input arguments to server
* it seems not good if each client like java, python has logic.

At least, power capacity value is checked by server side now. (it's unnecessary check by client side.)
I'll close this PR after checking set_power_ac behavior.

ki4070ma:
These checks are already done by server side. > set_power_ac, set_power_capacity
This PR is unnecessary.

mykola-mokhnach:
Yep, imagine the situation when a new value has been added or removed. Then it will be necessary to sync the value verification across all clients instead of doing that once on the server side. This is the reason people select languages like Python - you have more control on your input. In Java though, you'd be pretty limited by enum type and it would be necessary to change the client code in order to alter endpoint arguments.

ki4070ma:
Absolutely. Thanks for details

Issue #353, opened 2019-04-27
Support missing api


ki4070ma:
## The problem

There are some missing api supports on python client

Example: https://github.com/appium/python-client/pull/351

## How to extract missing api support

```
$cd /path/to/appium/commands-yml/commands
$find . -name "*yml" | while read FILE; do if [ `grep "python: false" $FILE | wc -l` = 1 ];then echo "****" && echo $FILE && grep python $FILE; fi; done
-> And checked if there are apis in python-client.
```

ki4070ma:
Missing apis

- [ ] GSM Signal
   - http://appium.io/docs/en/commands/device/network/gsm-signal/
- [ ] GSM Call
   - http://appium.io/docs/en/commands/device/network/gsm-call/
- [ ] GSM Voice
   - http://appium.io/docs/en/commands/device/network/gsm-voice/
- [ ] Toggle Data
   - http://appium.io/docs/en/commands/device/network/toggle-data/
- [ ] Toggle Airplane Mode
   - http://appium.io/docs/en/commands/device/network/toggle-airplane-mode/
- [ ] Get System Bars
   - http://appium.io/docs/en/commands/device/system/system-bars/
- [ ] Get Performance Data
   - http://appium.io/docs/en/commands/device/performance-data/get-performance-data/
- [ ] Get Performance Data Types
   - http://appium.io/docs/en/commands/device/performance-data/performance-data-types/
- [ ] Emulate power capacity
   - http://appium.io/docs/en/commands/device/emulator/power_capacity/
- [ ] Emulate power state
   - http://appium.io/docs/en/commands/device/emulator/power_ac/

PR #360, opened 2019-05-01
Support make_gsm_call

ki4070ma:
For https://github.com/appium/python-client/issues/353

Refer to https://github.com/appium/python-client/blob/master/appium/webdriver/applicationstate.py for refactoring.


appium/webdriver/gsm_call_actions.py
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env python
```
KazuCocoa: What about creating files in `gsm` directory like below?

```
appium/webdriver/extentions/gsm/call_actions.py
```

Or define the class in appium/webdriver/extensions/gsm.py

```
class Gsm(webdriver.Remote):
    ....

class GsmCallActions(object):
   ...
```

appium/webdriver/extensions/gsm.py
```diff
@@ -16,39 +16,50 @@
...
+        if strength not in self._dict_const(GsmSignalStrength).values():
+            logger.warning('{} is out of range. Use the value in {}. (e.g. GsmSignalStrength.GOOD)'.format(
+                strength, list(self._dict_const(GsmSignalStrength).keys())))
         self.execute(Command.SET_GSM_SIGNAL, {'signalStrength': strength, 'signalStrengh': strength})
         return self
 
-    def _dict_signal_strength(self):
-        return {'{}.{}'.format(Gsm.__name__, attr): value for attr, value in vars(Gsm).items()
-                if not callable(getattr(Gsm, attr)) and attr.isupper()}
+    def _dict_const(self, cls):
```
mykola-mokhnach: I think this should not be a class method, but rather a helper function. Can we also have a better name for it?

appium/webdriver/extensions/gsm.py
```diff
@@ -16,39 +16,50 @@
...
+
+        :Args:
+         - phone_number (str): The phone number to call to.
+         - action (str): The action - GsmCallActions.CALL/ACCEPT/CANCEL/HOLD
+
+        :Usage:
+            self.driver.make_gsm_call('5551234567', GsmCallActions.CALL)
+        """
+        if action not in self._dict_const(GsmCallActions).values():
+            logger.warning('{} is invalid. Use the value in {}. (e.g. GsmCallActions.CALL)'.format(
```
mykola-mokhnach: I'd rather say `is unknown` and `Consider using one of ... constants`

appium/webdriver/gsm_call_actions.py
ki4070ma: make sense, I did the latter one.

appium/common/helper.py
```diff
@@ -38,6 +35,19 @@ def appium_bytes(value, encoding):
         return value  # Python 2
 
 
+def get_dict_const_attr_and_val(cls):
```
ki4070ma: [want] better name (if you have)

mykola-mokhnach: how about `extract_items` or `extract_attributes` ?

ki4070ma: thanks, will rename to ```extract_const_attributes```.
(```isupper()``` works for const.)

appium/webdriver/extensions/gsm.py
```diff
@@ -46,7 +46,7 @@ def make_gsm_call(self, phone_number, action):
         :Usage:
             self.driver.make_gsm_call('5551234567', GsmCallActions.CALL)
         """
-        constants = get_dict_const_attr_and_val(GsmCallActions)
+        constants = extract_const_attributes(GsmCallActions)
         if action not in constants.values():
```
mykola-mokhnach: You need an ordered dict there or array of pairs, since the order of keys and values is not guaranteed to be stable in the default dict implementation

ki4070ma: Hum, could you tell me why ordered dict is necessary here?

> the order of keys and values is not guaranteed to be stable in the default dict implementation

I see this behavior.

mykola-mokhnach: https://stackoverflow.com/questions/39980323/are-dictionaries-ordered-in-python-3-6/39980744

mykola-mokhnach: means the actual behaviour depends on the Python version, so I wouldn't rely on it

ki4070ma: Thanks, I didn't know details from impl point of view.
I mean, what does it happen if dict order is not guaranteed?

My understanding is as below.

1. No impact to the results
   *  both ```'call' not in ['call', 'accept']``` and ```'call' not in ['accept', 'call']``` have the same result
2. Warning log can be changed along to python version
   * ```Consider using one of [CALL, ACCEPT]``` and ```Consider using one of [ACCEPT, CALL]``` can show up in user env.
   * If this order should be fixed, I'll fix dict order.

If my understanding is short, let me know.

mykola-mokhnach: for this particular case it does not matter, but for other things it might be good to keep the original order when showing the list of available options. You could simply use OrderedDict from `collections` to ensure such functionality without changing the client methods

appium/common/helper.py
```diff
@@ -39,13 +41,14 @@ def extract_const_attributes(cls):
     """
     Return dict with constants attributes and values in the class (e.g. {'VAL1': 1, 'VAL2': 2})
```
KazuCocoa: Is it better to address the order here?
_Return ordered dict with ..._

ki4070ma: At least IMO, it's unnecessary for users since not guaranteed order is due to dict behavior. (,,,but might update in next PR.)

PR #362, opened 2019-05-02
Resolve ci exception log on py37

ki4070ma:
* Exception log
   * https://travis-ci.org/appium/python-client/jobs/527252733
   * Actually, this exception log is no impact to results.

* Resolved exception log
   * https://travis-ci.org/appium/python-client/jobs/527311285

* How to resolve
   * Upgraded astroid and pylint for python3
   * Can't be upgraded for python2 since the latest astroid is unavailable on python2.

### DIFF
--- a/ci-requirements-py3.txt
+++ b/ci-requirements-py3.txt
@@ -1,0 +1,2 @@
+astroid==2.2.5
+pylint==2.3.1
--- a/ci-requirements-py2.txt
+++ b/ci-requirements-py2.txt
@@ -1,0 +1,2 @@
+astroid==1.6.5
+pylint==1.9.3
--- a/ci-requirements.txt
+++ b/ci-requirements.txt
@@ -1,7 +1,5 @@
 selenium==3.141.0
-astroid==1.6.5
 isort==4.3.4
-pylint==1.9.3
 autopep8==1.4.3
 httpretty==0.9.6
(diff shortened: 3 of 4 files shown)


ci-requirements-py3.txt
```diff
@@ -0,0 +1,2 @@
+astroid==2.2.5
+pylint==2.3.1
```
KazuCocoa: Can the version consider Python 2?

I wonder if pylint 2 claims errors which might break Python 2, but good for Python 3.

ki4070ma: > I wonder if pylint 2 claims errors which might break Python 2, but good for Python 3.

 I can't say it won't happen.
Your point is that we should have the same pylint version for python2/3, right?

Since exception log is no impact to the results, we can upgrade those version when we drop python2 support.

KazuCocoa: Yes.
We support both of them by one codebase.

---

btw, we have https://github.com/appium/python-client/blob/master/development.txt .
It refers to `ci-requirements.txt`. we should update here, too.

ki4070ma:
Thanks for your comments. I agree with you.

> btw, we have https://github.com/appium/python-client/blob/master/development.txt .
It refers to ci-requirements.txt. we should update here, too.

Thanks for this, too. I didn't care of it...

PR #368, opened 2019-05-08
Support get_performance_data, get_performance_data_types

ki4070ma:
For https://github.com/appium/python-client/issues/353


appium/webdriver/extensions/performance.py
```diff
@@ -0,0 +1,55 @@
...
+
+from selenium import webdriver
+from ..mobilecommand import MobileCommand as Command
+
+
+class Performance(webdriver.Remote):
+
+    def get_performance_data(self, package_name, data_type, data_read_timeout=None):
+        """Returns the information of the system state
+        which is supported to read as like cpu, memory, network traffic, and battery
```
KazuCocoa: Could you address about "For Android"? Because Python client has all of APIs in one module

Issue #369, opened 2019-05-09
Fix broken appium_service in Python 3


KazuCocoa:
## The problem
`import httplib` does not work for Python 3
https://python-future.org/compatible_idioms.html#http-module

https://discuss.appium.io/t/appium-service-throws-error-in-python-3/26328

PR #370, opened 2019-05-09
fix poll_url in Python 3

KazuCocoa:
closes https://github.com/appium/python-client/issues/369

`httplib` is available only in Python 2 environment.
SImply, I replaced the http request checking with `PoolManager`.

I tested locally in both Python 2 and 3 env.
Added a simple test case to ensure it can load in both python 2 and 3 environments.

### DIFF
--- a/appium/webdriver/appium_service.py
+++ b/appium/webdriver/appium_service.py
@@ -13,12 +13,11 @@
 # limitations under the License.
 
 
-import httplib
 import os
 import subprocess
 import sys
 import time
-
+import urllib3
 
 DEFAULT_HOST = '127.0.0.1'
 DEFAULT_PORT = 4723
@@ -49,9 +48,10 @@ def poll_url(host, port, path, timeout_ms):
     time_started_sec = time.time()
     while time.time() < time_started_sec + timeout_ms / 1000.0:
         try:
-            conn = httplib.HTTPConnection(host=host, port=port, timeout=1.0)
-            conn.request('HEAD', path)
-            if conn.getresponse().status < 400:
+            conn = urllib3.PoolManager(timeout=1.0)
+            resp = conn.request('HEAD', 'http://{host}:{port}{path}'.format(
+                host=host, port=port, path=path))
+            if resp.status < 400:
                 return True
         except Exception:
             pass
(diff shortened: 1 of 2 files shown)


johnkulovitz:
Thanks for doing this fix! I'm getting a new error, however. The join function used in this seems to be giving an error about mixing strings and bytes in path components. I'll put my pytest error output below, and I can give any extra information you may need!

`../../Library/Python/3.7/lib/python/site-packages/appium/webdriver/appium_service.py:158: in start
    main_script = kwargs['main_script'] if 'main_script' in kwargs else self._get_main_script()
../../Library/Python/3.7/lib/python/site-packages/appium/webdriver/appium_service.py:92: in _get_main_script
    if os.path.exists(os.path.join(modules_root, MAIN_SCRIPT_PATH)):
/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/posixpath.py:94: in join
    genericpath._check_arg_types('join', a, *p)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

funcname = 'join', args = (b'/usr/local/lib/node_modules', 'appium/build/lib/main.js'), hasstr = True, hasbytes = True, s = 'appium/build/lib/main.js'

    def _check_arg_types(funcname, *args):
        hasstr = hasbytes = False
        for s in args:
            if isinstance(s, str):
                hasstr = True
            elif isinstance(s, bytes):
                hasbytes = True
            else:
                raise TypeError('%s() argument must be str or bytes, not %r' %
                                (funcname, s.__class__.__name__)) from None
        if hasstr and hasbytes:
>           raise TypeError("Can't mix strings and bytes in path components") from None
E           TypeError: Can't mix strings and bytes in path components

/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/genericpath.py:151: TypeError
`

mykola-mokhnach:
@johnkulovitz Please report this as a separate bug

PR #372, opened 2019-05-09
Fix functional tests failed (android, ime/multi_action)

ki4070ma:
For https://github.com/appium/python-client/issues/373

## Results
Verified with Python2/3 and emulator(AndroidP) on https://github.com/appium/python-client/pull/372/commits/7c53fd31b4657932db7e55198efee3e02085f49f

* Python2/Python3

```
➜  android git:(fix-test-ime-multiaction) ✗ python2 ime_tests.py    
test_activate_ime_engine (__main__.IMETests) ... ok
test_active_ime_engine (__main__.IMETests) ... ok
test_available_ime_engines (__main__.IMETests) ... ok
test_deactivate_ime_engine (__main__.IMETests) ... ok
test_is_ime_active (__main__.IMETests) ... ok

----------------------------------------------------------------------
Ran 5 tests in 144.501s

OK
➜  android git:(fix-test-ime-multiaction) ✗ python3 ime_tests.py
test_activate_ime_engine (__main__.IMETests) ... ok
test_active_ime_engine (__main__.IMETests) ... ok
test_available_ime_engines (__main__.IMETests) ... ok
test_deactivate_ime_engine (__main__.IMETests) ... ok
test_is_ime_active (__main__.IMETests) ... ok

----------------------------------------------------------------------
Ran 5 tests in 131.105s

OK
➜  android git:(fix-test-ime-multiaction) ✗ python2 multi_action_tests.py 
test_actions_with_waits (__main__.MultiActionTests) ... ok
test_driver_multi_tap (__main__.MultiActionTests) ... ok
test_parallel_actions (__main__.MultiActionTests) ... ok

----------------------------------------------------------------------
Ran 3 tests in 115.003s

OK
➜  android git:(fix-test-ime-multiaction) ✗ python3 multi_action_tests.py
test_actions_with_waits (__main__.MultiActionTests) ... ok
test_driver_multi_tap (__main__.MultiActionTests) ... ok
test_parallel_actions (__main__.MultiActionTests) ... ok

----------------------------------------------------------------------
Ran 3 tests in 115.334s

OK

```

### DIFF
--- a/appium/webdriver/common/multi_action.py
+++ b/appium/webdriver/common/multi_action.py
@@ -18,6 +18,8 @@
 # These do not quite work for this situation, and do not allow for ad hoc action
 # chaining as the spec requires.
 
+import copy
+
 from appium.webdriver.mobilecommand import MobileCommand as Command
 
 
@@ -45,7 +47,12 @@ class MultiAction(object):
             if self._touch_actions is None:
                 self._touch_actions = []
 
-            self._touch_actions.append(touch_action)
+            try:
+                # deep copy, so that once they are in here, the user can't muck about
+                self._touch_actions.append(copy.deepcopy(touch_action))
+            except TypeError:
+                # For python3
+                self._touch_actions.append(copy.copy(touch_action))
 
     def perform(self):
         """Perform the actions stored in the object.
--- a/appium/webdriver/common/multi_action.py
+++ b/appium/webdriver/common/multi_action.py
@@ -18,8 +18,6 @@
 # These do not quite work for this situation, and do not allow for ad hoc action
 # chaining as the spec requires.
 
-import copy
-
 from appium.webdriver.mobilecommand import MobileCommand as Command
 
 
@@ -47,8 +45,7 @@ class MultiAction(object):
             if self._touch_actions is None:
                 self._touch_actions = []
 
-            # deep copy, so that once they are in here, the user can't muck about
-            self._touch_actions.append(copy.deepcopy(touch_action))
+            self._touch_actions.append(touch_action)
 
     def perform(self):
         """Perform the actions stored in the object.
--- a/appium/webdriver/common/multi_action.py
+++ b/appium/webdriver/common/multi_action.py
@@ -47,12 +47,7 @@ class MultiAction(object):
             if self._touch_actions is None:
                 self._touch_actions = []
 
-            try:
-                # deep copy, so that once they are in here, the user can't muck about
-                self._touch_actions.append(copy.deepcopy(touch_action))
-            except TypeError:
-                # For python3
-                self._touch_actions.append(copy.copy(touch_action))
+            self._touch_actions.append(copy.copy(touch_action))
 
     def perform(self):
         """Perform the actions stored in the object.
--- a/test/functional/android/helper/test_helper.py
+++ b/test/functional/android/helper/test_helper.py
@@ -12,16 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import sys
-
 from selenium.webdriver.support import expected_conditions as EC
 from selenium.webdriver.support.ui import WebDriverWait
 
 SLEEPY_TIME = 3
-
-
-def is_py3():
-    return sys.version_info.major == 3
 
 
 def wait_for_element(driver, locator, value, timeout=SLEEPY_TIME):
--- a/test/functional/android/helper/test_helper.py
+++ b/test/functional/android/helper/test_helper.py
@@ -12,10 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import sys
+
 from selenium.webdriver.support import expected_conditions as EC
 from selenium.webdriver.support.ui import WebDriverWait
 
 SLEEPY_TIME = 3
+
+
+def is_py3():
+    return sys.version_info.major == 3
 
 
 def wait_for_element(driver, locator, value, timeout=SLEEPY_TIME):
--- a/test/functional/android/ime_tests.py
+++ b/test/functional/android/ime_tests.py
@@ -23,7 +23,8 @@ import desired_capabilities
 # the emulator is sometimes slow and needs time to think
 SLEEPY_TIME = 1
 
-LATIN_IME = 'com.google.android.inputmethod.latin/com.android.inputmethod.latin.LatinIME'
+ANDROID_LATIN = 'com.android.inputmethod.latin/.LatinIME'  # Android L/M/N
+GOOGLE_LATIN = 'com.google.android.inputmethod.latin/com.android.inputmethod.latin.LatinIME'  # Android O/P
 
 
 class IMETests(unittest.TestCase):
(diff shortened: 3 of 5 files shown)


test/functional/android/helper/test_helper.py
```diff
@@ -12,12 +12,18 @@
...
 
+import sys
+
 from selenium.webdriver.support import expected_conditions as EC
 from selenium.webdriver.support.ui import WebDriverWait
 
 SLEEPY_TIME = 3
 
 
+def is_py3():
```
mykola-mokhnach: Can we rather rely on the behaviour instead of the version number?

test/functional/android/ime_tests.py
```diff
@@ -45,7 +44,8 @@ def test_is_ime_active(self):
         self.assertTrue(self.driver.is_ime_active())
 
     def test_active_ime_engine(self):
-        self.assertIsInstance(self.driver.active_ime_engine, unicode)
+        expected = str if is_py3() else unicode
+        self.assertIsInstance(self.driver.active_ime_engine, expected)
```
mykola-mokhnach:
```
try:
  basestring
except NameError:
  basestring = str

isinstance(s, basestring)
```

appium/webdriver/common/multi_action.py
```diff
@@ -47,8 +47,12 @@ def add(self, *touch_actions):
...
                 self._touch_actions = []
 
-            # deep copy, so that once they are in here, the user can't muck about
-            self._touch_actions.append(copy.deepcopy(touch_action))
+            try:
+                # deep copy, so that once they are in here, the user can't muck about
+                self._touch_actions.append(copy.deepcopy(touch_action))
+            except TypeError:
+                # For python3
+                self._touch_actions.append(copy.copy(touch_action))
```
mykola-mokhnach: is deepcopy not present in py3?

test/functional/android/ime_tests.py
ki4070ma: Thanks, it's cool.

appium/webdriver/common/multi_action.py
ki4070ma: No, py3 has deepcopy, but deepcopy doesn't work for own class(e.g. TouchAction) in py3.

https://stackoverflow.com/questions/32593035/why-does-deepcopy-fail-when-copying-a-complex-object

In that case, ```__deepcopy__``` has to be implemented by ourselves.

e.g. at ```TouchAction```
```
    def __deepcopy__(self, memodict={}):
        new_obj = TouchAction(self._driver)
        for action in self._actions:
            new_obj._add_action(action['action'], action['options'])
        return new_obj
```

At least, I have to consider current impl again since copy is not equal to deepcopy.
Steps are as below.

1. Check how current ```deepcopy``` works. If no problem, I'd like to remove it.
1. Implement own ```__deepcopy__```

test/functional/android/ime_tests.py
ki4070ma: ^ unfortunately, below exception occurs on py2, so current impl is revised.

```
UnboundLocalError: local variable 'basestring' referenced before assignment
```

appium/webdriver/common/multi_action.py
mykola-mokhnach: we just have to decide if shallow copy is enough for such case. In case it is not enough then I would rather suggest implementing __deepcopy__

test/functional/android/ime_tests.py
```diff
@@ -45,7 +43,11 @@ def test_is_ime_active(self):
         self.assertTrue(self.driver.is_ime_active())
 
     def test_active_ime_engine(self):
-        self.assertIsInstance(self.driver.active_ime_engine, unicode)
+        try:
+            expected = basestring
+        except NameError:
```
mykola-mokhnach: Please add a comment here about why this except is needed

test/functional/android/ime_tests.py
```diff
@@ -16,16 +16,14 @@
...
-
 from appium import webdriver
 import desired_capabilities
 
 
 # the emulator is sometimes slow and needs time to think
 SLEEPY_TIME = 1
 
-LATIN_IME = u'com.android.inputmethod.latin/.LatinIME'
+LATIN_IME = 'com.google.android.inputmethod.latin/com.android.inputmethod.latin.LatinIME'
```
KazuCocoa: here depends on Android OS versions

So, what about updateing to like below?

```python
ANDROID_LATIN = 'com.android.inputmethod.latin/.LatinIME'
GOOGLE_LATIN = 'com.google.android.inputmethod.latin/com.android.inputmethod.latin.LatinIME'

def test_available_ime_engines(self):
    engines = self.driver.available_ime_engines
    self.assertTrue(ANDROID_LATIN in engines or GOOGLE_LATIN in engines)
```

test/functional/android/ime_tests.py
```diff
@@ -45,7 +43,11 @@ def test_is_ime_active(self):
         self.assertTrue(self.driver.is_ime_active())
 
     def test_active_ime_engine(self):
-        self.assertIsInstance(self.driver.active_ime_engine, unicode)
+        try:
```
KazuCocoa: You also can modify like:

```
engines = self.driver.available_ime_engines
self.assertTrue(self.driver.active_ime_engine in engines)
```

instead of adding except.

appium/webdriver/common/multi_action.py
ki4070ma: Sorry for my missing check, but here ```deepcop``` doesn't work also in python2.
I've checked python2 behavior after adding try-catch.
With my previous codes, shallow copy worked in both py2 and py3.

### Summary
* Here ```deepcopy``` doesn't work both in py2 and py3..
   * py2: TBU
   * py3: https://stackoverflow.com/questions/32593035/why-does-deepcopy-fail-when-copying-a-complex-object
* Replaced ```deepcopy``` with ```copy```
   * Verified ```test_driver_multi_tap``` working
   * Verified ```test_driver_multi_tap``` working with objects with same value as below. Objects with same values are recognized as different one.

```
positions = [(100, 200), (100, 400), (100, 200)]
self.driver.tap(positions)
```

log by ```print(actions)``` at [json_wire_gestures@muti_action.py](https://github.com/appium/python-client/blame/129bed8f6f4e4c5e2145fde5f53a89fcd4845654/appium/webdriver/common/multi_action.py#L77)
```
[[{'action': 'press', 'options': {'x': 100, 'y': 200}}, {'action': 'release', 'options': {}}], [{'action': 'press', 'options': {'x': 100, 'y': 400}}, {'action': 'release', 'options': {}}], [{'action': 'press', 'options': {'x': 100, 'y': 200}}, {'action': 'release', 'options': {}}]]
```


Any comments to my checks are welcome.

test/functional/android/ime_tests.py
ki4070ma: Though I tried to keep current verification, your idea is better.

* (Before) Check if type is expected(```basestring```, ```str```)
* (After) Check if the value is within ```available_ime_engines```.

Issue #365, opened 2019-05-07
Move to Pipfile for dependencies management


mykola-mokhnach:
Currently the module dependencies are managed by `requirements.txt`. This is not good, since such format does not provide good compatibility with multiple python versions and also does not allow to properly use dependencies lock approach. We need to switch deps management to use [Pipfile](https://github.com/pypa/pipfile), which is actively developed and has much more possibilities for fine-tuning

KazuCocoa:
nice 👍

ki4070ma:
Just introduced pipfile by https://github.com/appium/python-client/pull/376.

We need to utilize pipfile function on demands in the future.
e.g. https://github.com/pypa/pipfile/blob/master/examples/Pipfile

PR #376, opened 2019-05-13
Introduced pipfile

ki4070ma:
For https://github.com/appium/python-client/issues/365


mykola-mokhnach:
Thanks Mori

Pipfile
```diff
@@ -0,0 +1,19 @@
...
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[dev-packages]
+pre-commit = "==1.13.0"
+
+[packages]
+selenium = "==3.141.0"
+astroid = "==1.6.5"
```
mykola-mokhnach: should we differentiate its version based on py3/py2?

Pipfile
```diff
@@ -0,0 +1,19 @@
+[[source]]
```
mykola-mokhnach: we can also explicitly point out which python versions are supported here

Pipfile
```diff
@@ -0,0 +1,19 @@
...
+[packages]
+selenium = "==3.141.0"
+astroid = "==1.6.5"
+isort = "==4.3.4"
+pylint = "==1.9.3"
+autopep8 = "==1.4.3"
+httpretty = "==0.9.6"
+pytest = "==4.0.2"
+pytest-cov = "==2.6.0"
+tox = "==3.6.0"
```
mykola-mokhnach: I would say it is fine to use `any minor version above this` rather than `equal to`

Pipfile
KazuCocoa: No, 1.6.5 is enough I think.
`astroid` is used by pylint. Pylint 1.9.3 is necessary to run it for both versions.
Once we drop Python 2.7, we can remove this deps and update the pylint, too.

mykola-mokhnach: Then it makes sense to keep a TODO comment there

tox.ini
```diff
@@ -8,7 +8,8 @@ envlist =
     py37
 
 [testenv]
-deps =
-    -r ci-requirements.txt
+deps = pipenv
```
ki4070ma: new line

Pipfile
ki4070ma: You mean `[requires]` section?

If so, I removed from below description.
If no and you have example, it would be helpful.

https://docs.pipenv.org/en/latest/basics/#general-recommendations-version-control
>Specify your target Python version in your Pipfile’s [requires] section. Ideally, you should only have one target Python version, as this is a deployment tool.

Pipfile
```diff
@@ -4,16 +4,18 @@ url = "https://pypi.org/simple"
...
-tox-travis = "==0.11"
+selenium = "~=3.141"
+autopep8 = "~=1.4"
+httpretty = "~=0.9"
+pytest = "~=4.0"
+pytest-cov = "~=2.6"
+tox = "~=3.6"
+tox-travis = "~=0.11"
+
+# TODO Update to the latest ver when py2 support dropped
```
KazuCocoa: thanks

mykola-mokhnach: I mean you could do something like `unittest2 = {version = ">=1.0,<3.0", markers="python_version < '2.7.9' or (python_version >= '3.0' and python_version < '3.4')"}`

Pipfile
mykola-mokhnach: yeah, sad but true: https://github.com/pypa/pipenv/issues/1050

Pipfile
ki4070ma: Yes, but at least for pylint/astroid/isort, point is https://github.com/appium/python-client/pull/362#discussion_r280427226 .

And as for other packages, same package versions are available for both py2 and py3.

PR #379, opened 2019-05-16
Add isort to pre-commit

ki4070ma:
It's to keep imports clean.

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -1,15 +1,12 @@
 #!/bin/bash
 
-result=$(python -m autopep8 -r --global-config .config-pep8 -d .)
-if [[ $result ]] ; then
-  echo $result
+
+if ! python -m autopep8 -r --global-config .config-pep8 -i . ; then
   echo "Please run command 'python -m autopep8 -r --global-config .config-pep8 -i .' on your local and commit the result"
   exit 1
 fi
 
-result=$(python -m isort -rc . | grep -v "Skipped")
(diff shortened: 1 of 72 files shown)


ci.sh
```diff
@@ -7,5 +7,12 @@ if [[ $result ]] ; then
   exit 1
 fi
 
+result=$(python -m isort -rc . | grep -v "Skipped")
+if [[ $result ]] ; then
```
mykola-mokhnach: what do we check here?

ki4070ma: Checking if unordered imports aren't remained

mykola-mokhnach: shouldn't we check the exit code instead (`$?`) ?

ki4070ma: I've checked its behavior at first, but unfortunately both cases(ordered, unordered) retrun `0`.

mykola-mokhnach: how about https://stackoverflow.com/questions/2480584/how-do-i-use-a-file-grep-comparison-inside-a-bash-if-else-statement ?

ki4070ma: Will look into again.
At least I found  below issue and codes.

https://github.com/timothycrosley/isort/issues/423
https://github.com/timothycrosley/isort/blob/d093f97c1d26854fbb082e43cfae8a062dacda56/isort/main.py#L382

ki4070ma: I could do as expected by using ` --check-only` opt for `isort` :)

PR #382, opened 2019-05-17
Assign w3c property on the command executor.

erustusagutu:
When we call `start_session()` on the webdriver we assign the w3c capabilities on the driver but not on the command_executor which then means that when we execute a command we do not remove the sessionId from the request which regresses from this change https://github.com/SeleniumHQ/selenium/commit/e6c2e7069c10852917846231439eeb807cd2e821.

### DIFF
--- a/appium/webdriver/webdriver.py
+++ b/appium/webdriver/webdriver.py
@@ -226,6 +226,7 @@ class WebDriver(
 
         # Double check to see if we have a W3C Compliant browser
         self.w3c = response.get('status') is None
+        self.command_executor.w3c = self.w3c
 
     def _merge_capabilities(self, capabilities):
         """


KazuCocoa:
Test failures should be green by https://github.com/appium/python-client/pull/383

erustusagutu:
@KazuCocoa When are you likely to release these changes when they get merged in? I'm not familiar with your release schedule.

KazuCocoa:
We have no regular release cycle. Just fixing bugs, adding new features, then I run release script.
(This will be shipped in a couple of days, I think)

erustusagutu:
Ok thanks.

KazuCocoa:
https://pypi.org/project/Appium-Python-Client/0.43/

PR #384, opened 2019-05-17
add assertions for command_executor.w3c

KazuCocoa:
- Add assertions for `command_executor.w3c`
- Fix a failing test


test/unit/webdriver/device/lock_test.py
```diff
@@ -47,7 +47,7 @@ def test_lock_no_args(self):
         driver.lock()
```
ki4070ma: Good to have type check (not point in this PR though)

https://github.com/appium/python-client/blob/cb39903ef8d33d985c7c52dd9ff218e6188b5e6a/test/unit/webdriver/device/location_test.py#L37

PR #386, opened 2019-05-19
Support set_network_speed

ki4070ma:
It's missing api support.

I've noticed by comparing below 2 files.
* https://github.com/appium/python-client/blob/master/appium/webdriver/mobilecommand.py
* https://github.com/appium/ruby_lib_core/blob/fd2cbde5466a144ce8dccfc6cf330b969bcc8fd3/lib/appium_lib_core/common/command/common.rb

### DIFF
--- a/appium/webdriver/extensions/android/network.py
+++ b/appium/webdriver/extensions/android/network.py
@@ -14,11 +14,13 @@
 
 from selenium import webdriver
 
+from appium.common.helper import extract_const_attributes
+from appium.common.logger import logger
 from appium.webdriver.mobilecommand import MobileCommand as Command
 
 
 class NetSpeed(object):
-    GSM = 'gsm'  # GSM/CSD (up: 14.4, down: 14.4)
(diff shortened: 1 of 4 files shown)


appium/webdriver/extensions/android/network.py
```diff
@@ -55,6 +69,25 @@ def toggle_wifi(self):
         self.execute(Command.TOGGLE_WIFI, {})
         return self
 
+    def set_network_speed(self, speed_type):
+        """Set the network speed emulation.
+        Android Emulator only.
+
+        :Args:
+         - speed_type (str): The network speed type.
+           A member of the const appium.webdriver.extensions.android.network.NetSpeed.
```
mykola-mokhnach: I think we could use better approach to make method cross-references: https://stackoverflow.com/questions/21289806/link-to-class-method-in-python-docstring

ki4070ma: Yes, we need to do so. 
(I did so only at this time..)
 https://github.com/appium/python-client/blob/af2fa7034fe3874e666316ce701ce25e8e64425b/appium/webdriver/extensions/device_time.py#L34

PR #388, opened 2019-05-21
Support get_display_density

ki4070ma:
It's missing api support.
(It's the last one for now, I would say.)

I've noticed by comparing below 2 files.
* https://github.com/appium/python-client/blob/master/appium/webdriver/mobilecommand.py
* https://github.com/appium/ruby_lib_core/blob/fd2cbde5466a144ce8dccfc6cf330b969bcc8fd3/lib/appium_lib_core/common/command/common.rb

### DIFF
--- a/appium/webdriver/extensions/android/display.py
+++ b/appium/webdriver/extensions/android/display.py
@@ -1,0 +1,31 @@
+#!/usr/bin/env python
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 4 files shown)


appium/webdriver/extensions/android/display.py
```diff
@@ -0,0 +1,34 @@
...
+from selenium import webdriver
+
+from appium.webdriver.mobilecommand import MobileCommand as Command
+
+
+class Display(webdriver.Remote):
+
+    def get_display_density(self):
+        """Get the display density, Android only
+
```
mykola-mokhnach: It would be useful here to describe the meaning of the returned units

Issue #325, opened 2019-02-04
Move commands in webdriver.py into other mixin's classes


KazuCocoa:
We'd like to reduce a number commands in webdriver.py, since the file has a bunch of commands in one file 

example: https://github.com/appium/python-client/pull/307

----

https://github.com/appium/python-client/blob/master/appium/webdriver/webdriver.py#L741

- LAUNCH_APP, CLOSE_APP
    - => https://github.com/appium/python-client/blob/master/appium/webdriver/extensions/applications.py
- GET_CURRENT_PACKAGE, END_TEST_COVERAGE, OPEN_NOTIFICATIONS
    - https://github.com/appium/python-client/blob/master/appium/webdriver/extensions/android/common.py (new file)
- SET_IMMEDIATE_VALUE, REPLACE_KEYS, CLEAR
    - https://github.com/appium/python-client/blob/master/appium/webdriver/extensions/keyboard.py
    - Maybe keyboard is enough since the three also type text

KazuCocoa:
updated

SGA-atsushi-mori:
I think it's done.
If no, please re-open.

ki4070ma:
I think it's done.
If no, please reopen.

PR #399, opened 2019-06-13
Moving reset method from WebDriver to Applications

VinayakaMayura:
## Change list

Move reset(self) method from webdriver.py to applications.py
 
## Types of changes
- [ ] No changes in production code.
- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Details

This is in reference to #325 . Moved the reset app method from Webdriver.py to applications.py

Already test is added for this method in app_test.py

### DIFF
--- a/appium/webdriver/extensions/applications.py
+++ b/appium/webdriver/extensions/applications.py
@@ -183,6 +183,12 @@ class Applications(webdriver.Remote):
             data['stringFile'] = string_file
         return self.execute(Command.GET_APP_STRINGS, data)['value']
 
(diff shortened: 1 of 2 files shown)


ki4070ma:
[Just memo] Although I re-ran CI and log was gone, this error is due to azure pipeline(like 'error occurs 
 due to disconnection with host'), I would say. So we can't do anything for that error in my understanding.

Issue #402, opened 2019-06-14
PyCharm does not suggest methods for Appium Python client


KazuCocoa:
Selenium WebDriver can.

e.g.

```
self.driver. find_element_by_accessibility_id('xxxx') # When `.` is inputted, PyCharm should suggest `click` etc
self.driver. find_element_by_xpath('xxxx') # When `.` is inputted, PyCharm suggests `click` etc. selenium webdriver has it
```

I'm not sure the logic of PyCharm yet

mykola-mokhnach:
This probably happens because of the mixin structure. In Python 3 it might work better as it gives a possibility to provide type hints

ki4070ma:
Seems working well on my environment, so it's not due to Appium Python client at least. (maybe PyCharm settings?)

![screenshot](https://user-images.githubusercontent.com/1549855/59545216-a7aa9300-8f45-11e9-9af9-0e59192a4f65.png)

ki4070ma:
Appium-Python-Client should be here.

![screenshot](https://user-images.githubusercontent.com/1549855/59545374-6e732280-8f47-11e9-9db5-c030547ccff9.png)

KazuCocoa:
What about `find_element_by_accessibility_id`? Not `by_id` @ki4070ma

ki4070ma:
@KazuCocoa 
Sorry for my missing check description for details, it doesn't work.

![スクリーンショット 2019-06-15 10 24 36](https://user-images.githubusercontent.com/1549855/59546461-0ed14300-8f58-11e9-9698-71c5ea35bd5c.png)

ki4070ma:
Fixed by https://github.com/appium/python-client/pull/404

AmanVerma1110:
i'm also facing same issue how to sortout pls help

PR #404, opened 2019-06-19
Add autocompletion for pycharm

ki4070ma:
For https://github.com/appium/python-client/issues/402

It's another topic which type of docstring is better. > pycharm typehinting, google, numpy.

### DIFF
--- a/appium/webdriver/webdriver.py
+++ b/appium/webdriver/webdriver.py
@@ -252,6 +252,8 @@ class WebDriver(
 
         Returns:
             `WebElement`
+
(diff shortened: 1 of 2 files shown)


appium/webdriver/webdriver.py
```diff
@@ -252,6 +252,8 @@ def find_element(self, by=By.ID, value=None):
 
         Returns:
             `WebElement`
+
+        :rtype: `MobileWebElement`
```
ki4070ma: [notice] If `WebElement`, it doesn't work. (import matter)

Issue #414, opened 2019-07-07
Error when I try to set location


dron4ik86:
## The problem
Get the error when I try to set location:
`self.driver.set_location(32.44432761840545, 34.897902299649895, None)`
`  File "/usr/local/lib/python3.7/site-packages/appium/webdriver/extensions/location.py", line 38, in set_location
    "altitude": float(altitude)
TypeError: float() argument must be a string or a number, not 'NoneType'`
Doesn't work on version 0.46 but on python-client 0.36 works fine for me.

## Environment
- Appium version (or git revision) that exhibits the issue: Appium 1.14.0
- Node.js version (unless using Appium.app|exe): v12.5.0
- Mobile platform/version under test: Android 8.1
- Real device or emulator/simulator: emulator
- Appium CLI or Appium.app|exe: CLI

ki4070ma:
`alt` should be optional as python-client. (need fix)
If necessary, set `0` to `alt` as workaround for now.
Its error is expected on the latest version.

KazuCocoa:
The 0.36 had `"altitude": str(altitude)`, but current one is `"altitude": float(altitude)`.
Accidentally, the previous one sent `None` to server. It worked since the `altitude` was ignored in Appium server side, in the native context.

Could you change your code as https://github.com/appium/python-client/issues/414#issuecomment-509001694 ?
Maybe, set the value as zero by default is better since it should not be optional in web context. (It is JSONWP tho. @ki4070ma )

dron4ik86:
Thanks!

PR #415, opened 2019-07-13
Change altitude optional as arg for set_location

ki4070ma:
## Changes
* Change altitude optional as arg for set_location

## Reasons
1. `altitude` is optional for appium server
   - http://appium.io/docs/en/commands/session/geolocation/set-geolocation/
      - JSON Parameters - altitude
2. It's enough that users set `altitude` when users want.

## Verified
Checked below No.1-3 cases work.

```
    self.driver.set_location(121.21, 11.56, 20)  # No.1
    # self.driver.set_location(121.21, 11.56)  # No.2
    # self.driver.set_location('121.22', '11.56', '21') # No.3
    import time
    time.sleep(10)
    print(self.driver.location)
```

Link to https://github.com/appium/python-client/issues/414

### DIFF
--- a/appium/webdriver/extensions/location.py
+++ b/appium/webdriver/extensions/location.py
@@ -35,7 +35,7 @@ class Location(webdriver.Remote):
         Args:
             latitude (float): String or numeric value between -90.0 and 90.00
             longitude (float): String or numeric value between -180.0 and 180.0
-            altitude (float, optional): String or numeric value
+            altitude (float, optional): String or numeric value (Android real device only)
 
         Returns:
             `appium.webdriver.webdriver.WebDriver`
--- a/appium/webdriver/extensions/location.py
+++ b/appium/webdriver/extensions/location.py
@@ -33,21 +33,21 @@ class Location(webdriver.Remote):
         """Set the location of the device
 
         Args:
-            latitude (float): String or numeric value between -90.0 and 90.00
-            longitude (float): String or numeric value between -180.0 and 180.0
-            altitude (float, optional): String or numeric value (Android real device only)
+            latitude (Union[float, str]): String or numeric value between -90.0 and 90.00
+            longitude (Union[float, str]): String or numeric value between -180.0 and 180.0
+            altitude (Union[float, str], optional): String or numeric value (Android real device only)
 
         Returns:
             `appium.webdriver.webdriver.WebDriver`
         """
         data = {
             "location": {
-                "latitude": float(latitude),
-                "longitude": float(longitude),
+                "latitude": latitude,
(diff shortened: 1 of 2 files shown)


appium/webdriver/extensions/location.py
```diff
@@ -29,13 +29,13 @@ def toggle_location_services(self):
...
 
-    def set_location(self, latitude, longitude, altitude):
+    def set_location(self, latitude, longitude, altitude=None):
         """Set the location of the device
 
         Args:
             latitude (float): String or numeric value between -90.0 and 90.00
             longitude (float): String or numeric value between -180.0 and 180.0
-            altitude (float): String or numeric value
+            altitude (float, optional): String or numeric value
```
ki4070ma: Add comments that `altitude` is available only for Android real device. (need check for ios)

ki4070ma: > need check for ios

Not found implementations for altitude by searching with `setGeoLocation`, `altitude` under appium organization.

appium/webdriver/extensions/location.py
```diff
@@ -44,9 +44,10 @@ def set_location(self, latitude, longitude, altitude):
             "location": {
                 "latitude": float(latitude),
                 "longitude": float(longitude),
-                "altitude": float(altitude)
             }
         }
+        if altitude:
```
mykola-mokhnach: is not None

appium/webdriver/extensions/location.py
```diff
@@ -44,9 +44,10 @@ def set_location(self, latitude, longitude, altitude):
...
+            data['location']['altitude'] = float(altitude)
```
mykola-mokhnach: please don't cast the values to float. the server code can properly handle both numbers and strings if they have parseable format

appium/webdriver/extensions/location.py
```diff
@@ -29,24 +29,25 @@ def toggle_location_services(self):
...
 
-    def set_location(self, latitude, longitude, altitude):
+    def set_location(self, latitude, longitude, altitude=None):
         """Set the location of the device
 
         Args:
-            latitude (float): String or numeric value between -90.0 and 90.00
-            longitude (float): String or numeric value between -180.0 and 180.0
-            altitude (float): String or numeric value
+            latitude (Union[float, str]): String or numeric value between -90.0 and 90.00
```
ki4070ma: `Union` doesn't work for type hinting on pycharm, but it's correct.

https://mypy.readthedocs.io/en/latest/kinds_of_types.html#union-types
https://www.python.org/dev/peps/pep-0484/#union-types

PR #420, opened 2019-07-25
add new method for zoom and pinch.

matthewdy:
add new method of zoom and pinch..

### DIFF
--- a/appium/webdriver/extensions/action_helpers.py
+++ b/appium/webdriver/extensions/action_helpers.py
@@ -149,3 +149,87 @@ class ActionHelpers(webdriver.Remote):
             .release()
         action.perform()
         return self
+
+    def zoom_element(self, element):
+        """Zooms in on an element a certain amount
+
+        :Args:
+         - element - the element to zoom
+
+        :Usage:
+            driver.zoom(element)
+        """
+        multi_touch = MultiAction()
+        size=element.size
+        position = element.location
+        coordinate_x=position.get('x')
+        coordinate_y=position.get('y')
+        width=size.get('width')
+        height=size.get('height')
+        center = {'x': coordinate_x+width/2, 'y': coordinate_y+height/2}
+        action1 = TouchAction(self)
+        action1.press(element).move_to(element, center.get('x'), center.get('y')-100).release()
+        action2 = TouchAction(self)
+        action2.press(element).move_to(element, center.get('x'), center.get('y')+100).release()
+        multi_touch.add(action1, action2).perform()
+        return self
+
+    def zoom_coordinate(self, x, y):
+        """Zooms in on an specified coordinate
+
+        :Args:
+             - x  x-coordinate to performance zoom in
+             - y  y-coordinate to performance zoom in
+
+        :Usage:
+            driver.zoom(x,y)
+        """
+        multi_touch = MultiAction()
+        action1 = TouchAction(self)
+        action1.press(x, y).move_to(x, y-100).release()
+        action2 = TouchAction(self)
+        action2.press(x, y).move_to(x, y+100).release()
+        multi_touch.add(action1, action2).perform()
+        return self
+
+    def pinch_element(self, element):
+        """pinch on an element a certain amount
+
+        :Args:
+         - element - the element to pinch on
+
+        :Usage:
+            driver.zoom(element)
+        """
+        multi_touch = MultiAction()
+        size = element.size
+        position = element.location
+        coordinate_x = position.get('x')
+        coordinate_y = position.get('y')
+        width = size.get('width')
+        height = size.get('height')
+        center = {'x': coordinate_x + width / 2, 'y': coordinate_y + height / 2}
+        action1 = TouchAction(self)
+        action1.press(element, center.get('x'), center.get('y')-100).move_to(element).release()
+        action2 = TouchAction(self)
+        action2.press(element, center.get('x'), center.get('y')+100).move_to(element).release()
+        multi_touch.add(action1, action2).perform()
+        return self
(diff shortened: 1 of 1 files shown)


appium/webdriver/extensions/action_helpers.py
```diff
@@ -149,3 +149,87 @@ def flick(self, start_x, start_y, end_x, end_y):
...
+        """Zooms in on an element a certain amount
+
+        :Args:
+         - element - the element to zoom
+
+        :Usage:
+            driver.zoom(element)
+        """
+        multi_touch = MultiAction()
+        size=element.size
```
KazuCocoa: What about element.rect?
If you consider W3C or MJSONWP, you can switch the process with `self.w3c` like https://github.com/appium/python-client/blob/2cfcf090faadd302674af066b3dce130bd3ff79b/appium/webdriver/extensions/action_helpers.py#L40

appium/webdriver/extensions/action_helpers.py
```diff
@@ -149,3 +149,87 @@ def flick(self, start_x, start_y, end_x, end_y):
...
+        position = element.location
+        coordinate_x=position.get('x')
```
KazuCocoa: We can expect they return x, y, with, height, so `position['x']` is preferred

appium/webdriver/extensions/action_helpers.py
```diff
@@ -149,3 +149,87 @@ def flick(self, start_x, start_y, end_x, end_y):
...
+        coordinate_y=position.get('y')
+        width=size.get('width')
+        height=size.get('height')
+        center = {'x': coordinate_x+width/2, 'y': coordinate_y+height/2}
+        action1 = TouchAction(self)
```
KazuCocoa: Can you consider to use W3C actions instead of TouchAction?

matthewdy: Thanks!  FYI: https://github.com/appium/java-client/commit/6b216815411a0d7ddf1679c1ceb0c10c5a779b3d#diff-f50beb490f1d493c61c0e8d7a963e5e8

matthewdy: @KazuCocoa Can you please let me know why we should use W3C actions ?

KazuCocoa: We have no exact plan when we drop MJSONWP so far, but we're going to focus on W3C protocol. Then, it will happen on the client side as well.

matthewdy: ok! let me do some changes~

appium/webdriver/extensions/action_helpers.py
ki4070ma: Could you follow the same format to existing docstring? Thanks.

[references]
https://github.com/appium/python-client/blob/57fafcb4c32a38d170fc8658fe017fce19d2ee44/appium/webdriver/webelement.py#L77-L88
https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

appium/webdriver/extensions/action_helpers.py
```diff
@@ -149,3 +149,87 @@ def flick(self, start_x, start_y, end_x, end_y):
             .release()
         action.perform()
         return self
+
+    def zoom_element(self, element):
```
ki4070ma: Could you follow the same format to existing docstring? Thanks.

[references]
https://github.com/appium/python-client/blob/57fafcb4c32a38d170fc8658fe017fce19d2ee44/appium/webdriver/webelement.py#L77-L88
https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

appium/webdriver/extensions/action_helpers.py
ki4070ma: Please rebase to the latest on master branch and run below commands.

```
$python -m autopep8 -a -r --global-config .config-pep8 -i .
```

appium/webdriver/extensions/action_helpers.py
matthewdy: @ki4070ma Sure.

appium/webdriver/extensions/action_helpers.py
matthewdy: @KazuCocoa  According to the page:http://appium.io/docs/en/commands/interactions/actions/
and https://seleniumhq.github.io/selenium/docs/api/java/org/openqa/selenium/interactions/Actions.html#tick-org.openqa.selenium.interactions.Action-
I used the method: drag_and_drop_by_offset() 
`        if self.w3c:
            actions = ActionChains(self)
            actions\
                .drag_and_drop_by_offset(element, 0, -100)
            actions\
                .drag_and_drop_by_offset(element, 0, 100)
            actions.perform()
        else:`
but it doesn't take effect, is there anything wrong ?

KazuCocoa: mm, could you provide me the appium log? I wonder what commands sent to Appium server.
For example, ruby lib implements the action like https://github.com/appium/ruby_lib_core/blob/master/test/functional/common_w3c_actions.rb#L21-L28 . I forgot if I tried such drag/drop method (and did not work well) tho...

matthewdy: Thanks.

manoj9788:
@KazuCocoa Are we waiting on something to get this added?

manoj9788:
@matthewdy Is it possible for you to work on this again?

matthewdy:
> @matthewdy Is it possible for you to work on this again?

I will take some time on this these days.

PR #426, opened 2019-08-17
Add screenrecord unittest

ki4070ma:
For https://github.com/appium/python-client/issues/283


appium/webdriver/extensions/screen_record.py
```diff
@@ -73,9 +73,7 @@ def start_recording_screen(self, **options):
                 but doing so results in larger movie files.
 
         Returns:
-            bytes: Base-64 encoded content of the recorded media file or an empty string
-                if the file has been successfully uploaded to a remote location
-                (depends on the actual `remotePath` value).
+            None
```
ki4070ma: `None` comes from http://appium.io/docs/en/commands/device/recording-screen/start-recording-screen/ .
However if `stop_screen_recording` isn't called after previous `start_screen_recording`, `bytes` will return.

https://github.com/appium/appium-android-driver/blob/1e019d6865db66d0ebcd70e4b198f0de39c3e632/lib/commands/recordscreen.js#L235

Sorry for missing my checks, I'll fix here.

test/unit/webdriver/screen_record_test.py
```diff
@@ -0,0 +1,48 @@
...
+class TestWebDriverScreenRecord(object):
+
+    @httpretty.activate
+    def test_start_recording_screen(self):
+        driver = android_w3c_driver()
+        httpretty.register_uri(
+            httpretty.POST,
+            appium_command('/session/1234567890/appium/start_recording_screen'),
+        )
+        assert driver.start_recording_screen(user='userA', password='12345') is None
```
ki4070ma: [memo] should be empty string

PR #427, opened 2019-08-19
Add events property

vrunoa:
Get the session to get events timing information. @KazuCocoa @jlipps let me know what you guys think on this approach. 

I'm seeing the branch name now "porperty" 🤦🏻‍♂️

PR #429, opened 2019-08-22
Add events property

dpgraham:
COPY OF https://github.com/appium/python-client/pull/427

Needed to run python isort

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -5,6 +5,7 @@ verify_ssl = true
 
 [dev-packages]
 pre-commit = "~=1.13"
+mock = "~=3.0"
 
 [packages]
 selenium = "~=3.141"
--- a/Pipfile
+++ b/Pipfile
@@ -5,7 +5,6 @@ verify_ssl = true
 
 [dev-packages]
 pre-commit = "~=1.13"
-mock = "~=3.0"
 
 [packages]
 selenium = "~=3.141"
@@ -20,6 +19,7 @@ tox-travis = "~=0.11"
 
 httpretty = "~=0.9"
 python-dateutil = "~=2.8"
+mock = "~=3.0"
 
 # TODO Update to the latest ver when py2 support dropped
 pylint = "~=1.9"
--- a/appium/version.py
+++ b/appium/version.py
@@ -1,1 +1,1 @@
-version = '0.46'
+version = '0.47'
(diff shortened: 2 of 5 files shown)


ki4070ma:
@dpgraham 
I've checked in my local so that `webdriver_test.py` needs fix.
(At least, `python -m isort -rc .` works in my local)

Please run below command.

```
$python -m isort -rc test/unit/webdriver/webdriver_test.py
```

dpgraham:
@ki4070ma I tried that already and it still fails the test

ki4070ma:
@dpgraham 
Add `mock = "~=3.0"` to `Pipfile`.

https://github.com/appium/python-client/blob/master/Pipfile

https://pypi.org/project/mock/

Error log from Travis
```
Traceback:

test/unit/webdriver/webdriver_test.py:22: in <module>

    from mock import patch

E   ImportError: No module named mock
```

Pipfile
```diff
@@ -5,6 +5,7 @@ verify_ssl = true
 
 [dev-packages]
 pre-commit = "~=1.13"
+mock = "~=3.0"
```
ki4070ma: Need to be under `[packages]`.

(Would like to re-consider which modules should be for dev-packages, packages)

KazuCocoa:
https://github.com/appium/python-client/commit/9e46aa117439341d0b25ced4fd7f9a9bcb7e4a10
mm, my change does not appear here.

`isort` seems to depend on environment.
By hand and by pre-commit hook on my end got a different result.
The former addressed https://github.com/appium/python-client/commit/9e46aa117439341d0b25ced4fd7f9a9bcb7e4a10 , but the later had no issue without the change.

ki4070ma:
@KazuCocoa 
Sorry for confusing you, here change(https://github.com/appium/python-client/pull/429/commits/f2a6abdbe09aad8daa6b31423ca19d48421a3f06) is from github.com directly, so pre-commit doesn't run.

KazuCocoa:
No.
I meant the result of `isort` by hand, `python -m isort -rc .`,  and by pre-commit had a different result on my local.
We should consider if we continue to use it or fix the dependency.

KazuCocoa:
https://pypi.org/project/Appium-Python-Client/0.47/
cc @vrunoa

PR #431, opened 2019-08-22
try run all scripts and exit 1 when something fails

KazuCocoa:


ci.sh
```diff
@@ -1,15 +1,37 @@
...
+  # install newer pylint for Python 3.7 environment on CI
+  if [[ $LINT_RESULT =~ "RuntimeError: generator raised StopIteration" ]] ; then
+    EXIT_STATUS=0
+  fi
+)
+
+(
+  python -m pytest test/unit/
+)
+if [[ $? -eq 1 ]] ; then
```
ki4070ma: `-ne 0` is safer.

http://doc.pytest.org/en/latest/usage.html

ci.sh
```diff
@@ -1,15 +1,37 @@
...
   echo "Please run command 'python -m isort -rc .' on your local and commit the result"
-  exit 1
+  EXIT_STATUS=1
 fi
 
-python -m pylint --rcfile .pylintrc appium test --errors-only
-python -m pytest test/unit/
+(
+  LINT_RESULT=$(python -m pylint --rcfile .pylintrc appium test --errors-only 2>&1 | tee /dev/tty)
+  if [[ $? -ne 0 ]] ; then
```
ki4070ma: I missed here, but I hope working for future as well.

https://github.com/PyCQA/pylint/blob/04cf3f8298fa1c718f950c90c34ee99a17f3c503/doc/user_guide/run.rst#L154

PR #432, opened 2019-08-26
Fix isort behavior for mock

ki4070ma:
I reproduced https://github.com/appium/python-client/pull/429#issuecomment-523845293 in my local. ( `mock` was handled as 1st party module.)
So this commit fixes isort behavior for `mock` as 3rd party module.


KazuCocoa:
Could you add a note in the development section in README to add the third party library as 'known_third_party'? It may help future contributors

PR #449, opened 2019-10-12
add toggle_airplane_mode api

manoj9788:


mykola-mokhnach:
This endpoint does not work on Android7+. Also, there is anyway set_network_connection method, which allows to do the same

PR #459, opened 2019-11-03
move commands from webdriver as mixins class

manoj9788:
per @KazuCocoa updated comment in #325

### DIFF
--- a/appium/webdriver/extensions/android/common.py
+++ b/appium/webdriver/extensions/android/common.py
@@ -1,0 +1,63 @@
+#!/usr/bin/env python
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from selenium import webdriver
+
+from appium.webdriver.mobilecommand import MobileCommand as Command
+
+
+class Common(webdriver.Remote):
+
+    def end_test_coverage(self, intent, path):
+        """Ends the coverage collection and pull the coverage.ec file from the device.
+
+        Android only.
+        See https://github.com/appium/appium/blob/master/docs/en/writing-running-appium/android/android-coverage.md
+
+        Args:
+            intent (str): description of operation to be performed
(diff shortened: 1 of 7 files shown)


appium/webdriver/extensions/keyboard.py
```diff
@@ -129,3 +129,9 @@ def _addCommands(self):
             ('POST', '/session/$sessionId/appium/device/press_keycode')
         self.command_executor._commands[Command.LONG_PRESS_KEYCODE] = \
             ('POST', '/session/$sessionId/appium/device/long_press_keycode')
+        self.command_executor._commands[Command.SET_IMMEDIATE_VALUE] = \
```
KazuCocoa: Ah...
SET_IMMEDIATE_VALUE, REPLACE_KEYS and CLEAR are referred by WebElement.py.
Could you move these three items into like `appium/webdriver/extensions/webelement.py` to separate them?
And LOCATION_IN_VIEW as well.

manoj9788: Okay, I just followed per your latest comment on the issue.
Will change the API's and tests again.
SET_IMMEDIATE_VALUE, REPLACE_KEYS, CLEAR and LOCATION_IN_VIEW to web element.py

appium/webdriver/extensions/android/common.py
```diff
@@ -0,0 +1,63 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from selenium import webdriver
+
+from appium.webdriver.mobilecommand import MobileCommand as Command
+
+
+class Common(webdriver.Remote):
```
mykola-mokhnach: I would rather keep these commands in the main driver class

appium/webdriver/extensions/keyboard.py
manoj9788: done! pls check

appium/webdriver/extensions/android/common.py
manoj9788: I've had a chat with @KazuCocoa on this and added per his recommendation.

mykola-mokhnach: I mean the term "Common" does not properly describe the purpose of all these commands, because, for example the `open_notifications` method only works for Android

manoj9788: I've exactly asked the same question during my chat with him - And that's why this common.py is under /extensions/android class

KazuCocoa: yeah, I thought only for Android was much better to put them in _android_ module than the main driver for all

manoj9788:
@KazuCocoa pls approve if it looks okay!

manoj9788:
Thanks!

Issue #336, opened 2019-02-18
Move/define search context in appium/webdriver/extensions/search_context.py


KazuCocoa:
## The problem

https://github.com/appium/python-client/pull/335/commits/6f23f89aaee0acc811f535144ea22c5790b0cb52#diff-9f2b90ea43eed700426cb52a45812bfdL90 is a good example. It removed some lines tho.

We define many commands in webdriver.py and webelement.py.
We're working to reduce the number of lines in them to make them simple.

For search context, we would like to move search context-related command from webdriver and webelement to extensions/search_context.

https://github.com/appium/python-client/pull/335/files also be an example. The PR was for data matcher.

_Move commands in webdriver.py into other mixin's classes #325_ is another issue about the mixin thing

jlipps:
Yes please!

ki4070ma:
Done by https://github.com/appium/python-client/pull/461

PR #461, opened 2019-11-03
Move search context methods from webdriver and webelement to search_context

ki4070ma:
For https://github.com/appium/python-client/issues/336


appium/webdriver/extensions/search_context/android.py
```diff
@@ -99,12 +87,70 @@ def _build_data_matcher(self, name=None, args=None, className=None):
...
+        Args:
+            uia_string: The element name in the Android UIAutomator library
+
+        Usage:
+            driver.find_element_by_android_uiautomator('.elements()[1].cells()[2]')
+
+        Returns:
+            `appium.webdriver.webelement.WebElement`
+
+        :rtype: `MobileWebElement`
```
ki4070ma: Need to check if type hinting works.

ki4070ma:
@KazuCocoa @mykola-mokhnach 
Can I have your review for package structure in this PR?

appium/webdriver/extensions/search_context/custom.py
```diff
@@ -0,0 +1,61 @@
...
+                "selector" is the string that will be passed to the custom element
+                finding plugin itself
+
+        Usage:
+            driver.find_elements_by_custom("foo:bar")
+
+        Returns:
+            :obj:`list` of :obj:`appium.webdriver.webelement.WebElement`
+
+        :rtype: list of `MobileWebElement`
```
mykola-mokhnach: why `rtype` is set for some methods and is not set for others?

appium/webdriver/extensions/search_context/search_context.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env python
```
mykola-mokhnach: the content of this file could be moved to `__init__.py`

mykola-mokhnach:
@ki4070ma Added some suggestions, but otherwise looks good

appium/webdriver/extensions/search_context/custom.py
ki4070ma: For auto completion https://github.com/appium/python-client/pull/404.
But I'll remove it if possible after check.

appium/webdriver/extensions/search_context/search_context.py
ki4070ma: Thanks, I didn't know that `__init__.py` can be used.

appium/webdriver/extensions/search_context/custom.py
ki4070ma: Unfortunately it's necessary for auto completion in IDE.

appium/webdriver/extensions/search_context/android.py
ki4070ma: It's necessary.

appium/webdriver/extensions/search_context/custom.py
mykola-mokhnach: I mean can we add rtype to other method declaration to make it consistent?

appium/webdriver/extensions/search_context/__init__.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env python
```
mykola-mokhnach: 👍

appium/webdriver/extensions/search_context/custom.py
ki4070ma: In my opinion, it's unnecessary. (it's confusing us though.)
Originally `rtype` is unnecessary since `Returns:` works for auto completion.

https://www.jetbrains.com/help/pycharm/type-syntax-for-docstrings.html
 - reStructuredText / Google / and so on

But auto completion doesn't work with `Returns:` for `find_element_xxx` somehow.
So added `rtype` to only `find_element_xxx`.

Will add one TODO to `rtype` in each class.

PR #462, opened 2019-11-04
Revert some changes to fix broken codes

ki4070ma:
Needs more changes to move commands to webelement, but for now just some changes are reverted.


KazuCocoa:
ah, they only have functional tests...
(I missed the pathname as unit and it was greened...)

manoj9788:
Cool, thanks @ki4070ma

PR #465, opened 2019-11-05
fix: Fix random disconnect issues due to missing retry policies

mykola-mokhnach:
This is expected to be a temporary solution until https://github.com/SeleniumHQ/selenium/pull/7746 is merged and published (which would probably never happen ;-) )

@haskellpointersauce Could you please try the patch with your test data?

### DIFF
--- a/appium/webdriver/appium_connection.py
+++ b/appium/webdriver/appium_connection.py
@@ -12,9 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import urllib3
-
-from selenium.webdriver.remote.remote_connection import RemoteConnection
+from selenium.webdriver.remote.remote_connection import RemoteConnection, urllib3
 
 from appium.common.helper import library_version
 
--- a/appium/webdriver/appium_connection.py
+++ b/appium/webdriver/appium_connection.py
@@ -12,10 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from selenium.webdriver.remote.remote_connection import RemoteConnection, urllib3
+from selenium.webdriver.remote.remote_connection import (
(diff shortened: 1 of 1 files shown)


appium/webdriver/appium_connection.py
```diff
@@ -12,17 +12,26 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from selenium.webdriver.remote.remote_connection import RemoteConnection
+from selenium.webdriver.remote.remote_connection import RemoteConnection, urllib3
```
KazuCocoa: btw, should we import `urllib3` from Selenium side instead of `import urllib3`?

mykola-mokhnach: yes. This is a third party library, which means that we should either explicitly include it into Pipfile (which might lead to version conflicts) or just use the one, which is bundled with Selenium (and that makes sense because we override methods of the class imported from Selenium)

KazuCocoa: 👍

mykola-mokhnach:
After further investigation we have some more ideas on what could be the reason of the unexpected connection termination. Will continue the investigation on it

haskellpointersauce:
I agree this no longer seems to be the solution we were looking for. Just in case someone googles this or something I thought I'd mention the following about what appears to be the retry policy utilized in the remote connection class (via defaults provided by urllib3):

Reading the docs for urllib3, it kinda seems like the default retry strategy is retries=None, which perhaps counterintuitively means "to retry until you receive a response." [1]

So if this is the case, this change would actually make the client perform fewer retries than would otherwise been performed.

[1] https://urllib3.readthedocs.io/en/latest/reference/index.html#module-urllib3.connectionpool

PR #469, opened 2019-11-09
Support for log_event and get_events command

ki4070ma:
For https://github.com/appium/appium/issues/13568

### DIFF
--- a/appium/webdriver/extensions/log_event.py
+++ b/appium/webdriver/extensions/log_event.py
@@ -19,9 +19,13 @@ from ..mobilecommand import MobileCommand as Command
 
 class LogEvent(webdriver.Remote):
 
-    def get_events(self):
+    def get_events(self, type=None):
         """ Retrieves events information from the current session
         (Since Appium 1.16.0)
+
+        Args:
+            type: (Not implemented yet)
+                https://github.com/appium/appium-base-driver/blob/master/lib/basedriver/commands/event.js
 
         Usage:
             events = driver.get_events()
@@ -33,7 +37,10 @@ class LogEvent(webdriver.Remote):
                     startTime: (int) Received time
                     endTime: (init) Response time
         """
-        return self.execute(Command.GET_EVENTS)['value']
+        data = {}
+        if type is not None:
+            data['type'] = type
+        return self.execute(Command.GET_EVENTS, data)['value']
 
     def log_event(self, vendor, event):
         """Log a custom event on the Appium server.
--- a/appium/webdriver/extensions/log_event.py
+++ b/appium/webdriver/extensions/log_event.py
@@ -19,13 +19,9 @@ from ..mobilecommand import MobileCommand as Command
 
 class LogEvent(webdriver.Remote):
 
-    def get_events(self, type=None):
+    def get_events(self):
         """ Retrieves events information from the current session
         (Since Appium 1.16.0)
-
-        Args:
-            type: (Not implemented yet)
-                https://github.com/appium/appium-base-driver/blob/master/lib/basedriver/commands/event.js
 
         Usage:
             events = driver.get_events()
@@ -37,10 +33,7 @@ class LogEvent(webdriver.Remote):
                     startTime: (int) Received time
                     endTime: (init) Response time
         """
-        data = {}
-        if type is not None:
-            data['type'] = type
-        return self.execute(Command.GET_EVENTS, data)['value']
+        return self.execute(Command.GET_EVENTS)['value']
 
     def log_event(self, vendor, event):
         """Log a custom event on the Appium server.
--- a/appium/webdriver/extensions/log_event.py
+++ b/appium/webdriver/extensions/log_event.py
@@ -19,13 +19,12 @@ from ..mobilecommand import MobileCommand as Command
 
 class LogEvent(webdriver.Remote):
 
-    @property
-    def events(self):
+    def get_events(self):
         """ Retrieves events information from the current session
         (Since Appium 1.16.0)
 
         Usage:
-            events = driver.events
+            events = driver.get_events()
 
         Returns:
             `dict`: A dictionary of events timing information containing the following entries
@@ -55,7 +54,7 @@ class LogEvent(webdriver.Remote):
 
             ```
             driver.log_event('appium', 'funEvent')
-            print(driver.events)
+            print(driver.get_events())
             ```
 
             produce the following output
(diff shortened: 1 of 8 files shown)


appium/webdriver/webdriver.py
```diff
@@ -413,14 +417,9 @@ def events(self):
             events = driver.events
 
         Returns:
-            `dict containing events timing information from the current session`
+            `dict`: containing events timing information from the current session`
```
mykola-mokhnach: we also need to mention the minimum supported appium version (1.16.0)

appium/webdriver/webdriver.py
```diff
@@ -440,6 +439,8 @@ def _addCommands(self):
             ('POST', '/session/$sessionId/touch/multi/perform')
         self.command_executor._commands[Command.SET_IMMEDIATE_VALUE] = \
             ('POST', '/session/$sessionId/appium/element/$id/value')
+        self.command_executor._commands[Command.GET_EVENTS] = \
```
mykola-mokhnach: there is also an endpoint to add events. I think it makes sense to extract these two into a separate mixin

appium/webdriver/webdriver.py
mykola-mokhnach: it makes sense to describe the output format

appium/webdriver/extensions/log_event.py
```diff
@@ -0,0 +1,82 @@
...
+
+        Usage:
+            events = driver.events
+
+        Returns:
+            `dict`: A dictionary of events timing information containing the following entries
+                commands: (`list` of `dict`) List of dictionaries containing the following entries
+                    cmd: (str) Sent command to appium server
+                    startTime: (int) Received time
+                    endTime: (init)Response time
```
mykola-mokhnach: typo in `init`?

appium/webdriver/extensions/log_event.py
```diff
@@ -0,0 +1,82 @@
...
+        """ Retrieves events information from the current session
+        (Since Appium 1.16.0)
+
+        Usage:
+            events = driver.events
+
+        Returns:
+            `dict`: A dictionary of events timing information containing the following entries
+                commands: (`list` of `dict`) List of dictionaries containing the following entries
+                    cmd: (str) Sent command to appium server
```
mykola-mokhnach: The command name that has been sent to the appium server

appium/webdriver/extensions/log_event.py
```diff
@@ -0,0 +1,82 @@
...
+                    endTime: (init)Response time
+        """
+        return self.execute(Command.GET_EVENTS)['value']
+
+    def log_event(self, vendor, event):
+        """Log a custom event on the Appium server.
+        (Since Appium 1.16.0)
+
+        Args:
+            vendor: The vendor to log
```
mykola-mokhnach: should we also mention types of the args?

appium/webdriver/extensions/log_event.py
```diff
@@ -0,0 +1,82 @@
...
+
+        Note:
+            When below commands are called,
+
+            ```
+            driver.log_event('appium', 'funEvent')
+            print(driver.events)
+            ```
+
+            output is as below.
```
mykola-mokhnach: The below command ... produces the following output ...

mykola-mokhnach:
cc @dpgraham 

Can you please take a look at this change? I remember you were working on https://github.com/appium/python-client/pull/429

mykola-mokhnach:
I am ok with the PR, although I would like @dpgraham to check it before merging, since it might be a breaking one

appium/webdriver/extensions/log_event.py
ki4070ma: Yes, thanks.

appium/webdriver/extensions/log_event.py
```diff
@@ -0,0 +1,82 @@
...
+
+from selenium import webdriver
+
+from ..mobilecommand import MobileCommand as Command
+
+
+class LogEvent(webdriver.Remote):
+
+    @property
+    def events(self):
```
KazuCocoa: https://github.com/appium/appium-base-driver/blob/master/lib/protocol/routes.js#L589-L594
`/events` route accept `type` argument as option.
It hasn't been implemented yet, but this `events` should not be property.

You can rename here as `get_events(self, type = None)` to keep another `events` as `session_capability['events']` as backword compatibility as https://github.com/appium/python-client/pull/469#issuecomment-552186783

ki4070ma: @KazuCocoa 
Thanks, I didn't understand for details, I will keep original `events` and create `get_events` as another one.

appium/webdriver/webdriver.py
```diff
@@ -422,8 +429,6 @@ def events(self):
             logger.warning('Could not find events information in the session. Error:', e)
             return {}
 
-    # pylint: disable=protected-access
```
ki4070ma: Restore

appium/webdriver/extensions/log_event.py
```diff
@@ -0,0 +1,81 @@
...
+            events = driver.get_events()
+
+        Returns:
+            `dict`: A dictionary of events timing information containing the following entries
+                commands: (`list` of `dict`) List of dictionaries containing the following entries
+                    cmd: (str) The command name that has been sent to the appium server
+                    startTime: (int) Received time
+                    endTime: (init) Response time
+        """
+        return self.execute(Command.GET_EVENTS)['value']
```
KazuCocoa: As a client, below two cases are available

`self.execute(Command.LOG_EVENT, {})`
`self.execute(Command.LOG_EVENT, {'type': nil})` (<= Appium server will handle this 'type' to filter events)

ki4070ma: Yes, I've checked and added, but removed by below commit since it isn't implemented yet.
In my opinion, it's enough to add `type` when it's ready.

https://github.com/appium/python-client/pull/469/commits/0a8406965cde76302cb1d19767d3722ad444de30

However, I'll restore `type` since users can use it as soon as server is ready, docstring is not ready though.

KazuCocoa: It could depend on server-side if the parameter works.
We can handle the behaviour only server-side modification (and documentation on appium.io), if clients follow [the route definition](https://github.com/appium/appium-base-driver/blob/master/lib/protocol/routes.js) like other commands.

KazuCocoa:
Added `type`.
Please check the description in https://github.com/appium/appium-base-driver/pull/368 about the available format

PR #470, opened 2019-11-19
Replace apk for functional test

ki4070ma:
https://github.com/appium/android-apidemos/releases/tag/v3.1.0

[Note]
* android sdk ver 28 cause test fails(SytemUI not responding dialog shows up.)


mykola-mokhnach:
> You can compress the apk as zip to reduce the size

`apk` is already a zip file, just with different extension ;)

KazuCocoa:
ah, yeah right. I meant it can compress the `.apk` more :P
(Current size in this pr is over 4MB)

mykola-mokhnach:
It only would be possible to compress the file better if no or very low compression factor was used during its creation, which is not the case with apk files.

KazuCocoa:
Yeah, true. But I was able to reduce from 4.8MB to 3.5MB in the maybe same test app in ruby lib core, so just commented :)

test/functional/android/helper/desired_capabilities.py
```diff
@@ -22,15 +22,17 @@ def PATH(p):
...
         'deviceName': 'Android Emulator',
-        'app': PATH('../../apps/{}'.format(app)),
         'newCommandTimeout': 240,
         'automationName': 'UIAutomator2',
         'uiautomator2ServerInstallTimeout': 120000,
         'adbExecTimeout': 120000
     }
 
+    if app is not None:
+        desired_caps['app'] = PATH('../../apps/{}'.format(app))
```
mykola-mokhnach: I prefer to use path.join instead of putting the path separator explicitly

test/functional/android/touch_action_tests.py
```diff
@@ -175,13 +175,13 @@ def test_drag_and_drop(self):
         el = wait_for_element(self.driver, MobileBy.ACCESSIBILITY_ID, 'Drag and Drop')
         action.tap(el).perform()
 
-        dd3 = wait_for_element(self.driver, MobileBy.ID, 'com.example.android.apis:id/drag_dot_3')
-        dd2 = self.driver.find_element_by_id('com.example.android.apis:id/drag_dot_2')
+        dd3 = wait_for_element(self.driver, MobileBy.ID, 'io.appium.android.apis:id/drag_dot_3')
```
mykola-mokhnach: It makes sense to extract the application identifier into a constant, so next time it is replaced it would not be necessary to search for all occurrences of it

ki4070ma: Thanks, exactly.

test/functional/android/activities_tests.py
```diff
@@ -24,14 +24,14 @@ def test_current_activity(self):
         self.assertEqual('.ApiDemos', activity)
 
     def test_start_activity_this_app(self):
-        self.driver.start_activity("com.example.android.apis", ".ApiDemos")
+        self.driver.start_activity("io.appium.android.apis", ".ApiDemos")
```
ki4070ma: APIDEMO_PKG_NAME

test/functional/android/touch_action_tests.py
ki4070ma: APIDEMO_PKG_NAME

test/unit/helper/test_helper.py
```diff
@@ -57,10 +57,10 @@ def android_w3c_driver():
                     'deviceName': 'emulator-5554',
                     'app': '/test/apps/ApiDemos-debug.apk',
                     'deviceUDID': 'emulator-5554',
-                    'appPackage': 'com.example.android.apis',
-                    'appWaitPackage': 'com.example.android.apis',
-                    'appActivity': 'com.example.android.apis.ApiDemos',
-                    'appWaitActivity': 'com.example.android.apis.ApiDemos'
+                    'appPackage': 'io.appium.android.apis',
```
ki4070ma: Keep this style for unit test.

PR #471, opened 2019-11-23
Move session/execute_mobile commands to mixin class

ki4070ma:
For https://github.com/appium/python-client/issues/325

### DIFF
--- a/appium/webdriver/extensions/execute_mobile_command.py
+++ b/appium/webdriver/extensions/execute_mobile_command.py
@@ -1,0 +1,61 @@
+#!/usr/bin/env python
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from selenium import webdriver
+
+
+class ExecuteMobileCommand(webdriver.Remote):
+
+    def press_button(self, button_name):
+        """Sends a physical button name to the device to simulate the user pressing.
+
+        iOS only.
+        Possible button names can be found in
+        https://github.com/appium/WebDriverAgent/blob/master/WebDriverAgentLib/Categories/XCUIDevice%2BFBHelpers.h
+
+        Args:
+            button_name (str): the button name to be sent to the device
+
+        Returns:
+            `appium.webdriver.webdriver.WebDriver`
+
+        """
+        data = {
+            'name': button_name
+        }
+        self.execute_script('mobile: pressButton', data)
+        return self
+
+    @property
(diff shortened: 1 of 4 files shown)


ki4070ma:
Current implementation seems working well..
I may change this PR to just for mixin.

KazuCocoa:
> Current implementation seems working well..

Do you mean it workes even though it has no `Command.GET_ALL_SESSIONS`?
I suppose selenium client has the command. (Maybe it will be removed in selenium 4)

ki4070ma:
@KazuCocoa 
Thanks for your info, I could find `getAllSessions` in selenium.

https://github.com/SeleniumHQ/selenium/blob/b300c358f65f33c0cf43177316f433601c027bdb/py/selenium/webdriver/remote/command.py#L35

(I referred java-client's code, I'm not sure it's correct though.
 https://github.com/appium/java-client/blob/fcf8581be05c4d67d76a5522dbfa1df9a9b07474/src/main/java/io/appium/java_client/MobileCommand.java#L284 )

KazuCocoa:
The command is only for mjsonwp, not in W3C WebDriver definition.
In some client such as Java and Ruby have different classes/definitions for each mode.

You can add the `GTE /session` in Appium code not to depend on selenium webdriver side change.

ki4070ma:
> The command is only for mjsonwp, not in W3C WebDriver definition.
> In some clients such as Java and Ruby have different classes/definitions for each mode.

Thanks, I'll check it for details.
https://github.com/appium/ruby_lib_core/tree/master/lib/appium_lib_core/common/command

> You can add the GTE /session in Appium code not to depend on selenium webdriver side change.
Thanks, I'll do that.

appium/webdriver/extensions/execute_mobile_command.py
```diff
@@ -0,0 +1,61 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from selenium import webdriver
+
+
+class ExecuteMobileCommand(webdriver.Remote):
```
ki4070ma: [Want] better naming.

Current name comes from http://appium.io/docs/en/commands/mobile-command/ .

mykola-mokhnach:
This command is only present in JSONWP spec, but not in W3C

appium/webdriver/extensions/session.py
```diff
@@ -0,0 +1,72 @@
...
+
+        Usage:
+            sessions = driver.all_sessions
+
+        Returns:
+            `dict`: containing all open sessions
+        """
+        return self.execute(Command.GET_ALL_SESSIONS)['value']
+
+    # pylint: disable=protected-access
```
ki4070ma: Delete.

ki4070ma:
@KazuCocoa @mykola-mokhnach 
Thanks for your comments.
Is it expected in this PR that below endpoint is added as mjsonwp, right?

https://github.com/appium/appium-base-driver/blob/master/docs/mjsonwp/protocol-methods.md
 - GET /wd/hub/sessions

ki4070ma:
@KazuCocoa 

Thanks for your explanations.
It's helpful for me to understand for details.

PR #477, opened 2019-12-21
[CI] Run with iOS 13.3 and Xcode 11.3

ki4070ma:


ki4070ma:
`FindByIOClassChainTests` also has problems for both below cases.

* test_find_element_by_path
* test_find_multiple_elements_by_path

I'll wait for appium 1.16 to be stable for a moment.

mykola-mokhnach:
> FindByIOClassChainTests also has problems for both below cases.

I'd rather say the page source got changes in iOS 13.3, so the same locator now matches different items. We should update the class chain locators to match the current page source in order to make the tests pass

test/functional/ios/search_context/find_by_ios_predicate_tests.py
```diff
@@ -35,6 +35,10 @@ def test_find_element_by_value(self):
         self.driver.find_element_by_ios_predicate('wdValue == "Buttons"')
 
     def test_find_element_by_isvisible(self):
+        if get_xcode_ver() == '11.3':
+            # TODO Remove skip when it's fixed
```
KazuCocoa: hm, interesting.
I checked the predicate on my local, then it worked expectedly. Xcode 11.3 and iOS 13.3. It might depend on Xcode/iOS side.

```
> @driver.find_elements :predicate, 'wdValue == "Buttons" AND visible == 0'
=> []
> @driver.find_element :predicate, 'wdValue == "Buttons" AND visible == 1'
=> #<Selenium::WebDriver::Element:0x..fa67af99212304fc id="32000000-0000-0000-A80D-010000000000">
```

ki4070ma: Thanks for your info.
I'll investigate more.

ki4070ma: [Notes]
Below environment works.
So as a result, its difference is client?...
Will look into communications between client and server.

### Environments
* iPhone 8, iOS 13.3 (Same to python's one)
* [Appium ver] 1.16.0-beta.3

### Commands and test codes

```
$bundle exec rake test:func:ios TESTS=test/functional/ios/ios/search_context_test.rb
```

```
def test_predicate
  e = @@driver.find_element :predicate, 'wdName == "Buttons" AND visible == 1'
  assert_equal 'Buttons', e.name
end
```

ki4070ma: Resolved by https://github.com/appium/appium/issues/13745#issuecomment-568625153

test/functional/ios/helper/test_helper.py
```diff
@@ -28,3 +29,14 @@ def setUp(self):
...
+
+
+def get_xcode_ver():
+    """Returns Xcode version
+
+    Returns:
+        str: Xcode version (e.g. '11.3')
+    """
+    lines = sp.check_output(['xcodebuild', '-version'])
+    return [line.decode('utf-8').split()[-1] for line in lines.strip().split(b'\n')
```
KazuCocoa: You can handle here like below.

```python
match = re.match(r'Xcode ([0-9\.]+)+', lines)
return None if match is None else match.group(1)
```

How to handle here if the index error occurred? Will called test case will fail because of index error?

test/functional/ios/keyboard_tests.py
```diff
@@ -46,9 +48,14 @@ def test_hide_keyboard_presskey_strategy(self):
         self.assertFalse(el.is_displayed())
 
     def test_hide_keyboard_no_key_name(self):
+        if get_xcode_ver() == '11.3':
+            # TODO Needs to dismiss dictation dialog.
```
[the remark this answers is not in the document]
ki4070ma: The dialog is here. `autoAcceptAlerts: True` didn't work.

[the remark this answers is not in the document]
KazuCocoa: Thanks.
What the _Enable Dictation_ switch in _settings -> keyboards_ ?

My simulators were disabled it by default, but I assume it was enabled by default on your env. Then, the dialogue appears before typing texts. If it was enabled, what will happen when you disable it once?

README.md
```diff
@@ -80,19 +80,19 @@ You also can run particular tests like below.
 ### Unit
 
 ```
-$ py.test test/unit
```
ki4070ma: [Notes] https://stackoverflow.com/questions/39495429/py-test-vs-pytest-command

test/functional/ios/keyboard_tests.py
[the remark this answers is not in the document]
ki4070ma: Thanks for your comments.

> What the Enable Dictation switch in settings -> keyboards ?
Unfortunately, I couldn't resolve by this in my local env since another dialog showed up. (Title: Enable 3rd party keyboard?)
But this behavior doesn't occur in ci env, so I'll resolve in my local env as another topic.

test/functional/ios/helper/test_helper.py
ki4070ma: Thanks for your comments.
I've removed here for now since it's unnecessary. (but I'll use `re` if possible for future.)

PR #479, opened 2019-12-29
chore: add try/catch in release script

KazuCocoa:
If the python environment has no `twine`, an exception will occur. The command only pushes the dist module to pypi. (Building the  module is in another  method)
It can run after other tasks, so let me show a message to push later when the command fails.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -8,6 +8,6 @@ if [ -z "$PYTHON_BIN_PATH" ]; then
 fi
 
 CONFIGURE_DIR=$(dirname "$0")
-"$PYTHON_BIN_PATH" "${CONFIGURE_DIR}/script/release.py" "$@"
+"$PYTHON_BIN_PATH" -m "${CONFIGURE_DIR}/script/release.py" "$@"
 
 echo "Finish release process"
--- a/release.sh
+++ b/release.sh
@@ -8,6 +8,6 @@ if [ -z "$PYTHON_BIN_PATH" ]; then
 fi
 
 CONFIGURE_DIR=$(dirname "$0")
-"$PYTHON_BIN_PATH" -m "${CONFIGURE_DIR}/script/release.py" "$@"
+"$PYTHON_BIN_PATH" "${CONFIGURE_DIR}/script/release.py" "$@"
 
 echo "Finish release process"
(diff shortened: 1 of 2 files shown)


script/release.py
```diff
@@ -65,7 +65,13 @@ def tag_and_generate_changelog(new_version_num):
 
 
 def upload_sdist(new_version_num):
-    call_bash_script('twine upload "dist/Appium-Python-Client-{}.tar.gz"'.format(new_version_num))
+    push_file = 'dist/Appium-Python-Client-{}.tar.gz'.format(new_version_num)
+    try:
+        call_bash_script('twine upload "{}"'.format(push_file))
+    except Exception as e:
+        print('Failed to upload {} to pypi. '
```
mykola-mokhnach: should the error be printed to stderr?
should we rethrow the exception after print?

release.sh
```diff
@@ -8,6 +8,6 @@ if [ -z "$PYTHON_BIN_PATH" ]; then
 fi
 
 CONFIGURE_DIR=$(dirname "$0")
-"$PYTHON_BIN_PATH" "${CONFIGURE_DIR}/script/release.py" "$@"
+"$PYTHON_BIN_PATH" -m "${CONFIGURE_DIR}/script/release.py" "$@"
```
ki4070ma: In my understanding, `-m` has no meaning.
Module name needs to follow after `-m`. (e.g. pip, pytest)

https://docs.python.org/3/using/cmdline.html

script/release.py
KazuCocoa: I thought it was enough to print messages as stdout without stop/rethrow errors.
Current usage is showing messages on CLI and should not stop this by this error.

release.sh
KazuCocoa: thanks. yeah, it's true

ki4070ma:
@KazuCocoa 

Just question

> If the python environment has no twine, 

In this case, doesn't this script exit during `validate_release_env()`?
Or `no twine` is different from `not installed twine` ?

KazuCocoa:
No, that was passed. I also expected the command stopped by it.
Yes. `twine` worked by `python -m twine` but did not only `twine` in `release.py`.
I assumed Python2.7/Python3/pyenv stuff... (Haven't dived deeply though)

PR #481, opened 2020-01-01
ci: Take screen record as evidence

ki4070ma:
[Notes]
* For both iOS and Android
* Stop taking screenshot


mykola-mokhnach:
btw, video recording on android emulators is only supported since API 27

Issue #468, opened 2019-11-09
Python 2 deprecation


mykola-mokhnach:
January 1st of 2020 is the EOL for Python2. I would say it makes sense to start working on python client version 2 (for us this is going to be `1.0.0`), which only supports python 3 (preferably Python 3.7+)

List of things I would like to see in the new major release (suggestion are highly welcome):
- Explicit types declarations (use `typing` module) -> High priority
- asyncio and async/await usage (for both network and file system i/o operations) -> Medium to Low priority, this might be a serious change and might require the users to update a lot of their code
- Proper exceptions inheritance  (use `raise from`) -> High priority
- Ordered dictionaries usage -> High priority
- Proper strings/bytearrays usage -> High priority
- Proper `super` usage -> High priority
- pytest usage -> Medium priority

ki4070ma:
As internal tasks

- [ ] Remove py2 from travis ci and run py38
   - https://github.com/appium/python-client/blob/master/.travis.yml
- [ ] Upgrade dependencies
   - https://github.com/appium/python-client/blob/master/Pipfile

manoj9788:
Thanks @mykola-mokhnach  Absolutely makes sense to start this work now!

ki4070ma:
Action sequence

https://github.com/appium/python-client/pull/478#issuecomment-570939200

mykola-mokhnach:
One more useful things which is present since Python 3.7 and which we could use: 

- [Data classes](https://realpython.com/python-data-classes/)

ki4070ma:
Might need to add `py.typed` file to repository. Will check later.

https://www.python.org/dev/peps/pep-0561/

KazuCocoa:
Python 2.7.18 came out

ki4070ma:
Done

PR #482, opened 2020-01-11
Add explicit type declarations

ki4070ma:
For https://github.com/appium/python-client/issues/468

### Remaining TODO after this PR
1. Look into `mypy.ini` for details

### DIFF
--- a/appium/common/helper.py
+++ b/appium/common/helper.py
@@ -17,7 +17,7 @@ from collections import OrderedDict
 from appium import version as appium_version
 
 
-def extract_const_attributes(cls) -> OrderedDict:
+def extract_const_attributes(cls: type) -> OrderedDict:
     """Return dict with constants attributes and values in the class(e.g. {'VAL1': 1, 'VAL2': 2})
 
     Args:
@@ -30,7 +30,7 @@ def extract_const_attributes(cls) -> OrderedDict:
         [(attr, value) for attr, value in vars(cls).items() if not callable(getattr(cls, attr)) and attr.isupper()])
 
 
-def library_version():
+def library_version() -> str:
     """Return a version of this python library
     """
 
--- a/appium/common/helper.py
+++ b/appium/common/helper.py
@@ -17,7 +17,7 @@ from collections import OrderedDict
 from appium import version as appium_version
 
 
-def appium_bytes(value, encoding):
+def appium_bytes(value: str, encoding: str) -> str:
     """Return a bytes-like object
 
     Has _appium_ prefix to avoid overriding built-in bytes.
--- a/appium/common/helper.py
+++ b/appium/common/helper.py
@@ -17,25 +17,6 @@ from collections import OrderedDict
 from appium import version as appium_version
 
 
-def appium_bytes(value: str, encoding: str) -> str:
-    """Return a bytes-like object
-
-    Has _appium_ prefix to avoid overriding built-in bytes.
-
-    Args:
-        value (str): A value to convert
-        encoding (str): A encoding which will convert to
-
-    Returns:
-        str: A bytes-like object
-    """
-
-    try:
-        return bytes(value, encoding)  # Python 3
-    except TypeError:
-        return value  # Python 2
-
-
 def extract_const_attributes(cls) -> OrderedDict:
     """Return dict with constants attributes and values in the class(e.g. {'VAL1': 1, 'VAL2': 2})
 
--- a/appium/common/helper.py
+++ b/appium/common/helper.py
@@ -36,7 +36,7 @@ def appium_bytes(value: str, encoding: str) -> str:
         return value  # Python 2
 
 
-def extract_const_attributes(cls):
+def extract_const_attributes(cls) -> OrderedDict:
     """Return dict with constants attributes and values in the class(e.g. {'VAL1': 1, 'VAL2': 2})
 
     Args:
--- a/appium/common/helper.py
+++ b/appium/common/helper.py
@@ -12,22 +12,22 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from collections import OrderedDict
+from typing import Dict
 
 from appium import version as appium_version
 
 
-def extract_const_attributes(cls: type) -> OrderedDict:
+def extract_const_attributes(cls: type) -> Dict:
     """Return dict with constants attributes and values in the class(e.g. {'VAL1': 1, 'VAL2': 2})
 
     Args:
         cls (type): Class to be extracted constants
 
     Returns:
-        OrderedDict: dict with constants attributes and values in the class
+        dict: dict with constants attributes and values in the class
     """
-    return OrderedDict(
-        [(attr, value) for attr, value in vars(cls).items() if not callable(getattr(cls, attr)) and attr.isupper()])
+    return dict([(attr, value) for attr, value in vars(cls).items()
+                 if not callable(getattr(cls, attr)) and attr.isupper()])
 
 
 def library_version() -> str:
--- a/appium/saucetestcase.py
+++ b/appium/saucetestcase.py
@@ -19,7 +19,7 @@ from __future__ import print_function
 import os
 import sys
 import unittest
-from typing import Any, List, TypeVar
+from typing import Any, Callable, List
 
 from sauceclient import SauceClient
 
@@ -30,8 +30,8 @@ SAUCE_ACCESS_KEY = os.environ.get('SAUCE_ACCESS_KEY')
 sauce = SauceClient(SAUCE_USERNAME, SAUCE_ACCESS_KEY)
 
 
-def on_platforms(platforms: List) -> Any:
-    def decorator(base_class: Any) -> None:
+def on_platforms(platforms: List[str]) -> Callable[[type], None]:
+    def decorator(base_class: type) -> None:
         module = sys.modules[base_class.__module__].__dict__
         for i, platform in enumerate(platforms):
             name = "%s_%s" % (base_class.__name__, i + 1)
--- a/appium/saucetestcase.py
+++ b/appium/saucetestcase.py
@@ -42,10 +42,10 @@ def on_platforms(platforms: List[str]) -> Callable[[type], None]:
 
 class SauceTestCase(unittest.TestCase):
     def setUp(self) -> None:
-        self.desired_capabilities['name'] = self.id()
+        self.desired_capabilities['name'] = self.id()  # type: ignore
         sauce_url = "http://%s:%s@ondemand.saucelabs.com:80/wd/hub"
         self.driver = webdriver.Remote(
-            desired_capabilities=self.desired_capabilities,
+            desired_capabilities=self.desired_capabilities,  # type: ignore
             command_executor=sauce_url % (SAUCE_USERNAME, SAUCE_ACCESS_KEY)
         )
         self.driver.implicitly_wait(30)
--- a/appium/saucetestcase.py
+++ b/appium/saucetestcase.py
@@ -19,6 +19,7 @@ from __future__ import print_function
 import os
 import sys
 import unittest
+from typing import Any, List, TypeVar
 
 from sauceclient import SauceClient
 
@@ -29,8 +30,8 @@ SAUCE_ACCESS_KEY = os.environ.get('SAUCE_ACCESS_KEY')
 sauce = SauceClient(SAUCE_USERNAME, SAUCE_ACCESS_KEY)
 
 
-def on_platforms(platforms):
-    def decorator(base_class):
+def on_platforms(platforms: List) -> Any:
+    def decorator(base_class: Any) -> None:
         module = sys.modules[base_class.__module__].__dict__
         for i, platform in enumerate(platforms):
             name = "%s_%s" % (base_class.__name__, i + 1)
@@ -40,7 +41,7 @@ def on_platforms(platforms):
 
 
 class SauceTestCase(unittest.TestCase):
-    def setUp(self):
+    def setUp(self) -> None:
         self.desired_capabilities['name'] = self.id()
         sauce_url = "http://%s:%s@ondemand.saucelabs.com:80/wd/hub"
         self.driver = webdriver.Remote(
@@ -49,7 +50,7 @@ class SauceTestCase(unittest.TestCase):
         )
         self.driver.implicitly_wait(30)
 
-    def tearDown(self):
+    def tearDown(self) -> None:
         print("Link to your job: https://saucelabs.com/jobs/%s" % self.driver.session_id)
         try:
             if sys.exc_info() == (None, None, None):
--- a/appium/webdriver/appium_connection.py
+++ b/appium/webdriver/appium_connection.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Dict
+from typing import Any, Dict
 
 from selenium.webdriver.remote.remote_connection import RemoteConnection
 
@@ -22,7 +22,7 @@ from appium.common.helper import library_version
 class AppiumConnection(RemoteConnection):
 
     @classmethod
-    def get_remote_connection_headers(cls, parsed_url: str, keep_alive: bool = True) -> Dict:
+    def get_remote_connection_headers(cls, parsed_url: str, keep_alive: bool = True) -> Dict[str, Any]:
         """Override get_remote_connection_headers in RemoteConnection"""
         headers = RemoteConnection.get_remote_connection_headers(parsed_url, keep_alive=keep_alive)
         headers['User-Agent'] = 'appium/python {} ({})'.format(library_version(), headers['User-Agent'])
--- a/appium/webdriver/appium_connection.py
+++ b/appium/webdriver/appium_connection.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from typing import Dict
+
 from selenium.webdriver.remote.remote_connection import RemoteConnection
 
 from appium.common.helper import library_version
@@ -20,7 +22,7 @@ from appium.common.helper import library_version
 class AppiumConnection(RemoteConnection):
 
     @classmethod
-    def get_remote_connection_headers(cls, parsed_url, keep_alive=True):
+    def get_remote_connection_headers(cls, parsed_url: str, keep_alive: bool = True) -> Dict:
         """Override get_remote_connection_headers in RemoteConnection"""
         headers = RemoteConnection.get_remote_connection_headers(parsed_url, keep_alive=keep_alive)
         headers['User-Agent'] = 'appium/python {} ({})'.format(library_version(), headers['User-Agent'])
--- a/appium/webdriver/appium_service.py
+++ b/appium/webdriver/appium_service.py
@@ -16,7 +16,7 @@ import os
 import subprocess as sp
 import sys
 import time
-from typing import Any, List, Optional, TypeVar
+from typing import Any, List, Optional, TypeVar, Union
 
 import urllib3
 
@@ -88,22 +88,22 @@ class AppiumService(object):
                                      'Make sure it is installed and present in PATH')
         return self._npm_executable
 
-    def _get_main_script(self) -> str:
+    def _get_main_script(self) -> Union[str, bytes]:
         if not hasattr(self, '_main_script'):
             for args in [['root', '-g'], ['root']]:
                 try:
                     modules_root = sp.check_output([self._get_npm()] + args).strip().decode('utf-8')
                     if os.path.exists(os.path.join(modules_root, MAIN_SCRIPT_PATH)):
-                        self._main_script = os.path.join(modules_root, MAIN_SCRIPT_PATH)
+                        self._main_script: Union[str, bytes] = os.path.join(modules_root, MAIN_SCRIPT_PATH)
                         break
                 except sp.CalledProcessError:
                     continue
             if not hasattr(self, '_main_script'):
                 try:
-                    self._main_script = str(sp.check_output(
+                    self._main_script = sp.check_output(
                         [self._get_node(),
                          '-e',
-                         'console.log(require.resolve("{}"))'.format(MAIN_SCRIPT_PATH)]).strip())
+                         'console.log(require.resolve("{}"))'.format(MAIN_SCRIPT_PATH)]).strip()
                 except sp.CalledProcessError as e:
                     raise AppiumServiceError(e.output)
         return self._main_script
@@ -190,8 +190,8 @@ class AppiumService(object):
             bool: `True` if the service was running before being stopped
         """
         is_terminated = False
-        if self.is_running and self._process is not None:
-            self._process.terminate()
+        if self.is_running:
+            self._process.terminate()  # type: ignore
             is_terminated = True
         self._process = None
         self._cmd = None
--- a/appium/webdriver/appium_service.py
+++ b/appium/webdriver/appium_service.py
@@ -16,7 +16,7 @@ import os
 import subprocess as sp
 import sys
 import time
-from typing import Any, Optional, Union
+from typing import Any, List, Optional, TypeVar
 
 import urllib3
 
@@ -45,7 +45,7 @@ def find_executable(executable: str) -> Optional[str]:
     return None
 
 
-def poll_url(host: str, port: str, path: str, timeout_ms: int) -> bool:
+def poll_url(host: str, port: int, path: str, timeout_ms: int) -> bool:
     time_started_sec = time.time()
     while time.time() < time_started_sec + timeout_ms / 1000.0:
         try:
@@ -64,12 +64,15 @@ class AppiumServiceError(RuntimeError):
     pass
 
 
+T = TypeVar('T', bound='AppiumService')
+
+
 class AppiumService(object):
     def __init__(self) -> None:
-        self._process = None
-        self._cmd = None
+        self._process: Optional[sp.Popen] = None
+        self._cmd: Optional[List] = None
 
-    def _get_node(self) -> Optional[str]:
+    def _get_node(self) -> str:
         if not hasattr(self, '_node_executable'):
             self._node_executable = find_executable('node')
         if self._node_executable is None:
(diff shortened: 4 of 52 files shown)


appium/common/helper.py
```diff
@@ -17,7 +17,7 @@
 from appium import version as appium_version
 
 
-def appium_bytes(value, encoding):
+def appium_bytes(value: str, encoding: str) -> str:
```
mykola-mokhnach: the method returns `bytes`

appium/saucetestcase.py
```diff
@@ -29,8 +30,8 @@
 sauce = SauceClient(SAUCE_USERNAME, SAUCE_ACCESS_KEY)
 
 
-def on_platforms(platforms):
-    def decorator(base_class):
+def on_platforms(platforms: List) -> Any:
```
mykola-mokhnach: this returns Callable

mykola-mokhnach: List[str]

appium/saucetestcase.py
```diff
@@ -29,8 +30,8 @@
...
+    def decorator(base_class: Any) -> None:
```
mykola-mokhnach: Can we have a more specific type instead of Any for a class argument?

appium/webdriver/appium_service.py
```diff
@@ -27,7 +27,7 @@
 STATUS_URL = '/wd/hub/status'
 
 
-def find_executable(executable):
+def find_executable(executable: str) -> Union[str, None]:
```
mykola-mokhnach: could also be Optional[str]

appium/webdriver/appium_service.py
```diff
@@ -119,7 +119,7 @@ def _parse_host(args):
                 return args[idx + 1]
         return DEFAULT_HOST
 
-    def start(self, **kwargs):
+    def start(self, **kwargs: Union[Dict, str, int]) -> sp.Popen:
```
mykola-mokhnach: to me it does not make much sense to add typing to kwargs

appium/webdriver/common/multi_action.py
```diff
@@ -68,7 +76,7 @@ def perform(self):
         return self
 
     @property
-    def json_wire_gestures(self):
+    def json_wire_gestures(self) -> Dict:
```
mykola-mokhnach: Dict could also be parametrised with key/value types, for example Dict[str, str]

appium/webdriver/common/touch_action.py
```diff
@@ -24,16 +24,23 @@
 # pylint: disable=no-self-use
 
 import copy
+from typing import Dict, List, TypeVar
 
 from appium.webdriver.mobilecommand import MobileCommand as Command
 
+# from appium.webdriver.webelement import WebElement as WebElement
```
mykola-mokhnach: Are these comments needed?

appium/webdriver/common/multi_action.py
```diff
@@ -19,17 +19,25 @@
 # chaining as the spec requires.
 
 import copy
+from typing import Dict, List, Optional, TypeVar
 
 from appium.webdriver.mobilecommand import MobileCommand as Command
 
+if False:  # For mypy
```
mykola-mokhnach: is that a recommended approach?

appium/webdriver/common/touch_action.py
```diff
@@ -24,16 +24,23 @@
...
 
 class TouchAction(object):
+
     def __init__(self, driver=None):
         self._driver = driver
-        self._actions = []
+        self._actions: List = []
 
-    def tap(self, element=None, x=None, y=None, count=1):
+    def tap(self: T, element=None, x: int = None, y: int = None, count: int = 1) -> T:
```
mykola-mokhnach: not sure if we need to also set the type of `self` arg

appium/webdriver/extensions/android/activities.py
```diff
@@ -76,6 +80,9 @@ def wait_activity(self, activity, timeout, interval=1):
             activity (str): target activity
             timeout (int): max wait time, in seconds
             interval (int): sleep interval between retries, in seconds
+
+        Returns:
+            bool: `True` if target activity shows
```
mykola-mokhnach: ...if the target activity is shown

appium/webdriver/extensions/android/common.py
```diff
@@ -48,12 +52,12 @@ def open_notifications(self):
         return self
 
     @property
-    def current_package(self):
+    def current_package(self) -> Any:  # TODO Check return type
```
mykola-mokhnach: it's a string

appium/webdriver/extensions/android/network.py
```diff
@@ -34,15 +38,15 @@ class NetSpeed(object):
 class Network(webdriver.Remote):
 
     @property
-    def network_connection(self):
+    def network_connection(self) -> Any:  # TODO Check return type
```
mykola-mokhnach: int

appium/webdriver/extensions/execute_mobile_command.py
```diff
@@ -38,7 +38,7 @@ def press_button(self, button_name):
         return self
 
     @property
-    def battery_info(self):
+    def battery_info(self) -> dict:
```
mykola-mokhnach: Dict

appium/webdriver/extensions/search_context/android.py
```diff
@@ -15,16 +15,21 @@
...
+if False:
+    from appium.webdriver.webelement import WebElement  # For mypy
+
 
 class AndroidSearchContext(BaseSearchContext):
     """Define search context for Android"""
 
-    def find_element_by_android_data_matcher(self, name=None, args=None, className=None):
+    def find_element_by_android_data_matcher(
+            self, name: str = None, args: str = None, className: str = None):  # -> WebElement:
```
mykola-mokhnach: why the retuned type is commented out?

appium/saucetestcase.py
mykola-mokhnach: https://stackoverflow.com/questions/44664040/type-hints-with-user-defined-classes/44664064

ki4070ma:
@mykola-mokhnach 
Thanks so much for your comments.
It's on-going, but I'm catching up explicit types declarations and your comments are very helpful.
I'll continue to proceed this PR.

appium/common/helper.py
ki4070ma: Thanks, I've removed `appium_bytes` since it's for the compatibility for py2 and py3.

appium/saucetestcase.py
ki4070ma: Yes, it's just temporal. But I'll check your link, thanks.

appium/webdriver/appium_service.py
ki4070ma: Yes, thanks.

appium/webdriver/appium_service.py
ki4070ma: Hum, it makes sense, thanks.
(It's unnecessary to maintain return type when new type can be taken by `**kwarg`.

appium/webdriver/common/touch_action.py
ki4070ma: No, it's just to avoid cyclic import to run unit test. I'll handle it later.
Same for others.

appium/webdriver/common/touch_action.py
ki4070ma: It's necessary.

https://mypy.readthedocs.io/en/stable/generics.html#generic-methods-and-generic-self

appium/webdriver/extensions/search_context/android.py
ki4070ma: Because unit test is failed due to cyclic import.
Will fix later.

appium/webdriver/common/multi_action.py
ki4070ma: No.
Now changed to `if TYPE_CHECKING`.

appium/webdriver/appium_connection.py
```diff
@@ -20,7 +22,7 @@
 class AppiumConnection(RemoteConnection):
 
     @classmethod
-    def get_remote_connection_headers(cls, parsed_url, keep_alive=True):
+    def get_remote_connection_headers(cls, parsed_url: str, keep_alive: bool = True) -> Dict:
```
ki4070ma: Will look into each `Dict` case which can be like `Dict[str, str]` later.

appium/webdriver/extensions/android/network.py
```diff
@@ -34,15 +38,15 @@ class NetSpeed(object):
...
+    def network_connection(self) -> int:
         """Returns an integer bitmask specifying the network connection type.
 
         Android only.
         Possible values are available through the enumeration `appium.webdriver.ConnectionType`
         """
         return self.execute(Command.GET_NETWORK_CONNECTION, {})['value']
 
-    def set_network_connection(self, connection_type):
+    def set_network_connection(self, connection_type: int) -> Any:  # TODO Check return type
```
KazuCocoa: Perhaps None ( or `''`)

appium/common/helper.py
```diff
@@ -17,26 +17,7 @@
 from appium import version as appium_version
 
 
-def appium_bytes(value, encoding):
```
KazuCocoa: 👋

appium/common/helper.py
```diff
@@ -17,26 +17,7 @@
...
-    """
-
-    try:
-        return bytes(value, encoding)  # Python 3
-    except TypeError:
-        return value  # Python 2
-
-
-def extract_const_attributes(cls):
+def extract_const_attributes(cls: type) -> OrderedDict:
```
mykola-mokhnach: we could use a normal dict now, since it is ordered by default

appium/webdriver/appium_service.py
```diff
@@ -64,62 +64,65 @@ class AppiumServiceError(RuntimeError):
...
                     if os.path.exists(os.path.join(modules_root, MAIN_SCRIPT_PATH)):
                         self._main_script = os.path.join(modules_root, MAIN_SCRIPT_PATH)
                         break
-                except subprocess.CalledProcessError:
+                except sp.CalledProcessError:
                     continue
             if not hasattr(self, '_main_script'):
                 try:
-                    self._main_script = subprocess.check_output(
+                    self._main_script = str(sp.check_output(
```
mykola-mokhnach: why it is necessary to cast to str here?

appium/webdriver/appium_service.py
```diff
@@ -64,62 +64,65 @@ class AppiumServiceError(RuntimeError):
...
     @staticmethod
-    def _parse_host(args):
+    def _parse_host(args: List[str]) -> str:
         for idx, arg in enumerate(args or []):
             if arg in ('--address', '-a') and idx < len(args) - 1:
                 return args[idx + 1]
         return DEFAULT_HOST
 
-    def start(self, **kwargs):
+    def start(self, **kwargs: Any) -> sp.Popen:
```
mykola-mokhnach: we don't have to set kwargs type

appium/webdriver/appium_service.py
```diff
@@ -187,15 +190,15 @@ def stop(self):
             bool: `True` if the service was running before being stopped
         """
         is_terminated = False
-        if self.is_running:
+        if self.is_running and self._process is not None:
```
mykola-mokhnach: `self._process ` is already checked inside `is_running`

appium/webdriver/common/multi_action.py
```diff
@@ -19,17 +19,25 @@
 # chaining as the spec requires.
 
 import copy
+from typing import TYPE_CHECKING, Dict, List, TypeVar, Union
 
 from appium.webdriver.mobilecommand import MobileCommand as Command
 
+if TYPE_CHECKING:
```
mykola-mokhnach: 👍

appium/webdriver/common/touch_action.py
```diff
@@ -24,16 +24,24 @@
...
 class TouchAction(object):
-    def __init__(self, driver=None):
+
+    def __init__(self, driver: 'WebDriver' = None):
         self._driver = driver
-        self._actions = []
+        self._actions: List = []
 
-    def tap(self, element=None, x=None, y=None, count=1):
+    def tap(self: T, element: 'WebElement' = None, x: int = None, y: int = None, count: int = 1) -> T:
```
mykola-mokhnach: Do we need to set the type of `self` argument?

appium/webdriver/common/touch_action.py
```diff
@@ -126,12 +134,14 @@ def release(self):
...
 
-    def perform(self):
+    def perform(self: T) -> T:
         """Perform the action by sending the commands to the server to be operated upon
 
         Returns:
             `TouchAction`: self instance
         """
+        if self._driver is None:
+            raise AttributeError('Set driver to constructor as a argument when to create the instance.')
```
mykola-mokhnach: I would rather verify if `self._driver` is not None when passing it to the constructor. Also https://stackoverflow.com/questions/8297526/proper-exception-to-raise-if-none-encountered-as-argument makes sense

appium/webdriver/extensions/action_helpers.py
```diff
@@ -61,7 +66,7 @@ def drag_and_drop(self, origin_el, destination_el):
         action.long_press(origin_el).move_to(destination_el).release().perform()
         return self
 
-    def tap(self, positions, duration=None):
+    def tap(self, positions: List[Tuple], duration: int = None) -> T:
```
mykola-mokhnach: Optional[int]

appium/webdriver/extensions/android/activities.py
```diff
@@ -12,15 +12,19 @@
...
 from selenium.webdriver.support.ui import WebDriverWait
 
 from appium.webdriver.mobilecommand import MobileCommand as Command
 
+T = TypeVar('T', bound='Activities')
+
 
 class Activities(webdriver.Remote):
-    def start_activity(self, app_package, app_activity, **opts):
+    def start_activity(self, app_package: str, app_activity: str, **opts: str) -> T:
```
mykola-mokhnach: `opts` can be passed without type

appium/webdriver/extensions/android/nativekey.py
```diff
@@ -1,3 +1,18 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
 class AndroidKey:
     # Key code constant: Unknown key code.
     UNKNOWN = 0
```
mykola-mokhnach: Python now (since v 3.4) implements `enum` [natively](https://docs.python.org/3/library/enum.html). We could think about implementing it for some existing classes

appium/webdriver/extensions/applications.py
```diff
@@ -47,7 +51,7 @@ def is_app_installed(self, bundle_id):
         }
         return self.execute(Command.IS_APP_INSTALLED, data)['value']
 
-    def install_app(self, app_path, **options):
+    def install_app(self, app_path: str, **options: Any) -> T:
```
mykola-mokhnach: `options` don't need a type

mykola-mokhnach: same below

appium/webdriver/extensions/images_comparison.py
```diff
@@ -12,14 +12,16 @@
...
+
 from selenium import webdriver
 
 from ..mobilecommand import MobileCommand as Command
 
 
 class ImagesComparison(webdriver.Remote):
 
-    def match_images_features(self, base64_image1, base64_image2, **opts):
+    def match_images_features(self, base64_image1: bytes, base64_image2: bytes, **opts: Any) -> Dict[str, Any]:
```
mykola-mokhnach: same here

appium/webdriver/extensions/images_comparison.py
```diff
@@ -73,7 +75,8 @@ def match_images_features(self, base64_image1, base64_image2, **opts):
         }
         return self.execute(Command.COMPARE_IMAGES, options)['value']
 
-    def find_image_occurrence(self, base64_full_image, base64_partial_image, **opts):
+    def find_image_occurrence(self, base64_full_image: bytes, base64_partial_image: bytes,
+                              **opts: Any) -> Tuple[bytes, Dict[str, int]]:
```
mykola-mokhnach: are you sure the returned type is a Tuple? I thought it is a dictionary

mykola-mokhnach: https://github.com/appium/appium-base-driver/blob/5b12eebcf011cf7676a2987b646a6394ddafb639/lib/basedriver/commands/images.js#L27

appium/webdriver/extensions/android/network.py
mykola-mokhnach: It returns the connection state after the change: https://github.com/appium/appium-android-driver/blob/64cd063862e7126c59a9f28d101250c08e3f1858/lib/commands/network.js#L91

appium/webdriver/extensions/location.py
```diff
@@ -29,7 +33,8 @@ def toggle_location_services(self):
         self.execute(Command.TOGGLE_LOCATION_SERVICES, {})
         return self
 
-    def set_location(self, latitude, longitude, altitude=None):
+    def set_location(self, latitude: Union[float, str], longitude: Union[float,
```
mykola-mokhnach: I'd rather split by arguments

appium/webdriver/extensions/log_event.py
```diff
@@ -35,14 +39,14 @@ def get_events(self, type=None):
                 commands: (`list` of `dict`) List of dictionaries containing the following entries
                     cmd: (str) The command name that has been sent to the appium server
                     startTime: (int) Received time
-                    endTime: (init) Response time
+                    endTime: (int) Response time
```
mykola-mokhnach: :)

appium/webdriver/extensions/remote_fs.py
```diff
@@ -13,15 +13,18 @@
...
 from selenium.common.exceptions import InvalidArgumentException
 
 from ..mobilecommand import MobileCommand as Command
 
+T = TypeVar('T', bound='RemoteFS')
+
 
 class RemoteFS(webdriver.Remote):
-    def pull_file(self, path):
+    def pull_file(self, path: str) -> bytes:
```
mykola-mokhnach: I think this endpoint returns base64-encoded string

appium/webdriver/extensions/search_context/android.py
```diff
@@ -15,16 +15,21 @@
...
+if TYPE_CHECKING:
+    from appium.webdriver.webelement import WebElement
+
 
 class AndroidSearchContext(BaseSearchContext):
     """Define search context for Android"""
 
-    def find_element_by_android_data_matcher(self, name=None, args=None, className=None):
+    def find_element_by_android_data_matcher(
+            self, name: str = None, args: str = None, className: str = None) -> 'WebElement':
```
mykola-mokhnach: Optional

mykola-mokhnach: same below

appium/webdriver/extensions/search_context/base_search_context.py
```diff
@@ -14,12 +14,17 @@
...
+
+if TYPE_CHECKING:
+    from appium.webdriver.webelement import WebElement
+
 
 class BaseSearchContext(object):
     """Used by each search context. Dummy find_element/s are for preventing pylint error"""
 
-    def find_element(self, by=None, value=None):
+    def find_element(self, by: str = None, value: Union[str, Dict] = None) -> 'WebElement':
```
mykola-mokhnach: str is Optional

appium/webdriver/extensions/session.py
```diff
@@ -33,19 +35,19 @@ def session(self):
...
 
         Returns:
-            `dict`: containing all open sessions
+            :obj:`list` of :obj:`dict`: containing all open sessions
         """
         return self.execute(Command.GET_ALL_SESSIONS)['value']
 
     @property
-    def events(self):
+    def events(self) -> Any:
```
mykola-mokhnach: Dict?

appium/webdriver/webdriver.py
```diff
@@ -141,8 +145,8 @@ class WebDriver(
     SystemBars
 ):
 
-    def __init__(self, command_executor='http://127.0.0.1:4444/wd/hub',
-                 desired_capabilities=None, browser_profile=None, proxy=None, keep_alive=True, direct_connection=False):
+    def __init__(self, command_executor: str = 'http://127.0.0.1:4444/wd/hub',
+                 desired_capabilities: Dict = None, browser_profile: str = None, proxy: str = None, keep_alive: bool = True, direct_connection: bool = False):
```
mykola-mokhnach: Optinonal

appium/webdriver/webelement.py
```diff
@@ -58,22 +56,19 @@ def get_attribute(self, name):
         if attributeValue is None:
             return None
 
-        if not isinstance(attributeValue, str):
```
mykola-mokhnach: I would say it is still needed

appium/webdriver/extensions/android/network.py
ki4070ma: Thanks, it seems `the connection state` as return val also from my checking actual behavior.

appium/webdriver/appium_service.py
ki4070ma: It was due to mypy error `appium/webdriver/appium_service.py:103: error: Incompatible types in assignment (expression has type "bytes", variable has type "str")` without cast.
But as you see, it changed current behavior and should be like below in definition.

```
self._main_script: Union[str, bytes] = os.path.join(modules_root, MAIN_SCRIPT_PATH)
```

appium/webdriver/appium_service.py
ki4070ma: Hum, it has below error without `Any` .

```
appium/webdriver/appium_service.py:125: error: Function is missing a type annotation for one or more arguments
```

And I couldn't find good solution from my search.
Do you have any idea? Or can I use `type: ignore`?

[ref]
* https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html

appium/webdriver/appium_service.py
ki4070ma: Thanks, it was due to below error.

```
appium/webdriver/appium_service.py:194: error: Item "None" of "Optional[Popen[Any]]" has no attribute "terminate"
```

Will add `type: ignore`.
(I guess it's mypy error.)

appium/webdriver/common/touch_action.py
ki4070ma: Yes, in my understanding.

https://github.com/appium/python-client/pull/482#discussion_r365553785

appium/webdriver/common/touch_action.py
ki4070ma: > I would rather verify if self._driver is not None when passing it to the constructor.
Some functions in `TouchAction` don't need `driver`, so its check cannot be moved to the constructor.

> Also https://stackoverflow.com/questions/8297526/proper-exception-to-raise-if-none-encountered-as-argument makes sense
Thanks.

appium/webdriver/extensions/android/nativekey.py
ki4070ma: Will think in another PR.

appium/webdriver/appium_service.py
mykola-mokhnach: ok, lets keep [Any](https://stackoverflow.com/questions/37031928/type-annotations-for-args-and-kwargs) for now

appium/webdriver/common/touch_action.py
mykola-mokhnach: makes sense

appium/webdriver/extensions/android/activities.py
ki4070ma: same as https://github.com/appium/python-client/pull/482#discussion_r365553785

appium/webdriver/extensions/session.py
ki4070ma: Thanks.

https://github.com/appium/python-client/blob/e16ab705a218b302dec9c61560af44c835051692/test/unit/webdriver/webdriver_test.py#L231

appium/webdriver/common/touch_action.py
```diff
@@ -24,16 +24,24 @@
 # pylint: disable=no-self-use
 
 import copy
+from typing import TYPE_CHECKING, Dict, List, TypeVar, Union
 
 from appium.webdriver.mobilecommand import MobileCommand as Command
 
+if TYPE_CHECKING:
+    from appium.webdriver.webelement import WebElement as WebElement
```
ki4070ma: Remove `as WebElement`

appium/webdriver/webelement.py
ki4070ma: Here is only for python2 since `unicode` is only in python2, so it's unnecessary.

mykola-mokhnach: not really. This is to make the spec happy, which says that attribute value could either be a string or null, but not boolean, number or something else

appium/webdriver/extensions/images_comparison.py
ki4070ma: Same as https://github.com/appium/python-client/pull/482#discussion_r365553785

appium/webdriver/webelement.py
ki4070ma: Ah, thanks, I haven't understood codes' meaning.
I restored and modified to `str` from `unicode`.

mykola-mokhnach: Please add an explanation comment to this block of the code, so it would be more clear to other people, who'd read it

appium/webdriver/webdriver.py
```diff
@@ -252,7 +258,7 @@ def _merge_capabilities(self, capabilities):
         w3c_caps = _make_w3c_caps(capabilities)
         return {'capabilities': w3c_caps, 'desiredCapabilities': capabilities}
 
-    def find_element(self, by=By.ID, value=None):
+    def find_element(self, by: Optional[str] = By.ID, value: Union[str, Dict] = None) -> MobileWebElement:
```
mykola-mokhnach: is `by` optional?

ki4070ma: Thanks, will revert.

PR #485, opened 2020-01-23
Add mypy to pre-commit

ki4070ma:
## Background
To avoid uploading codes with mypy error

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,3 +8,10 @@
     hooks:
     -   id: isort
         args: ["-rc", "."]
+-   repo: https://github.com/pre-commit/mirrors-mypy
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -8,3 +8,10 @@
     hooks:
     -   id: isort
         args: ["-rc", "."]
+-   repo: https://github.com/pre-commit/mirrors-mypy
+    rev: 'v0.761'
+    hooks:
+    -   id: mypy
```
ki4070ma: [Notes]
It didn't work by `args: ["appium"]`.

https://github.com/pre-commit/mirrors-mypy/issues/5#issuecomment-515841288

PR #486, opened 2020-01-25
chore: Applied some py3 formats

ki4070ma:
### Changes
* Used `f'This is {item}'` instead of `'This is {}'.format(item)` for some cases.
* Removed unnecessary `(object)` from class definitions.
* Removed unused imports

### DIFF
--- a/appium/webdriver/appium_service.py
+++ b/appium/webdriver/appium_service.py
@@ -102,7 +102,7 @@ class AppiumService:
                     self._main_script = sp.check_output(
                         [self._get_node(),
                          '-e',
-                         f'console.log(require.resolve("{MAIN_SCRIPT_PATH}"))']).strip()
+                         'console.log(require.resolve("{}"))'.format(MAIN_SCRIPT_PATH)]).strip()
                 except sp.CalledProcessError as e:
                     raise AppiumServiceError(e.output)
         return self._main_script
--- a/appium/webdriver/appium_service.py
+++ b/appium/webdriver/appium_service.py
@@ -50,8 +50,7 @@ def poll_url(host: str, port: int, path: str, timeout_ms: int) -> bool:
     while time.time() < time_started_sec + timeout_ms / 1000.0:
         try:
             conn = urllib3.PoolManager(timeout=1.0)
-            resp = conn.request('HEAD', 'http://{host}:{port}{path}'.format(
-                host=host, port=port, path=path))
+            resp = conn.request('HEAD', f'http://{host}:{port}{path}')
             if resp.status < 400:
                 return True
         except Exception:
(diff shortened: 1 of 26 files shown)


appium/webdriver/appium_service.py
```diff
@@ -103,7 +102,7 @@ def _get_main_script(self) -> Union[str, bytes]:
                     self._main_script = sp.check_output(
                         [self._get_node(),
                          '-e',
-                         'console.log(require.resolve("{}"))'.format(MAIN_SCRIPT_PATH)]).strip()
```
ki4070ma: Revert here. (Previous one is better.)

appium/webdriver/webdriver.py
```diff
@@ -185,20 +185,15 @@ def _update_command_executor(self, keep_alive: bool) -> None:
         if (not {direct_protocol, direct_host, direct_port, direct_path}.issubset(set(self.capabilities))):
             message = 'Direct connect capabilities from server were:\n'
             for key in [direct_protocol, direct_host, direct_port, direct_path]:
-                message += '{}: \'{}\'\n'.format(key, self.capabilities.get(key, ''))
+                message += f'{key}: \'{self.capabilities.get(key, "")}\'\n'
```
ki4070ma: Revert here. (Previous one is better.)

appium/webdriver/connectiontype.py
```diff
@@ -26,7 +26,7 @@
 """
 
 
-class ConnectionType(object):
+class ConnectionType:
```
ki4070ma: [Notes]
Couldn't use `enum` here since `ConnectionType(Enum)` will be different type from below return value.

https://github.com/appium/python-client/blob/e16ab705a218b302dec9c61560af44c835051692/appium/webdriver/extensions/android/network.py#L69

Need to change such like `return CoonectionType(self.execute(Command.SET_NETWORK_CONNECTION, data)['value'])`, but this breaks backward compatibility.

appium/webdriver/extensions/remote_fs.py
```diff
@@ -74,7 +74,7 @@ def push_file(self, destination_path: str,
                 with open(source_path, 'rb') as f:
                     file_data = f.read()
             except IOError:
-                message = 'source_path {} could not be found. Are you sure the file exists?'.format(source_path)
+                message = f'source_path {source_path} could not be found. Are you sure the file exists?'
```
mykola-mokhnach: I'd prefer to quote the actual values to properly read them in case they contain spaces

test/unit/webdriver/webdriver_test.py
```diff
@@ -52,7 +52,7 @@ def test_create_session(self):
 
         request = httpretty.HTTPretty.latest_requests[0]
         assert request.headers['content-type'] == 'application/json;charset=UTF-8'
-        assert 'appium/python {} (selenium'.format(appium_version.version) in request.headers['user-agent']
+        assert f'appium/python {appium_version.version} (selenium' in request.headers['user-agent']
```
ki4070ma: Revert. Previous one is better.

PR #487, opened 2020-01-25
chore: Fix mypy errors under test folder

ki4070ma:


test/functional/ios/helper/desired_capabilities.py
```diff
@@ -12,20 +12,21 @@
...
     os.path.join(os.path.dirname(__file__), p)
 )
 
 
 BUNDLE_ID = 'com.example.apple-samplecode.UICatalog'
 
 
-def get_desired_capabilities(app=None):
-    desired_caps = {
+def get_desired_capabilities(app: Optional[str] = None) -> Dict[str, Union[str, bool, int]]:
```
mykola-mokhnach: Dict[str, Any] would be easier to read

ki4070ma: I agree, and it will be more correct.

PR #515, opened 2020-04-11
feat: Override send_keys without file upload function

KazuCocoa:
Will update the naming, but let me create a PR as a draft

background: https://github.com/appium/appium-desktop/issues/1335
`send_keys` by some selenium bindings send a local file binary if the given string is a file.
But https://github.com/appium/appium-desktop/issues/1335 has a use case won't send a local file even if the given string is a local path.

### DIFF
--- a/appium/webdriver/webelement.py
+++ b/appium/webdriver/webelement.py
@@ -206,15 +206,17 @@ class WebElement(AppiumWebElementSearchContext):
         self._execute(Command.SET_IMMEDIATE_VALUE, data)
         return self
 
-    def send_keys_direct(self, *value):
+    # Override
+    def send_keys(self, *value):
         """Simulates typing into the element.
 
-        Please consider to use this method if WebElement#send_keys did not work.
-        WebElement#send_keys works uploading a local file, but this method simply
-        sends given string without such uploading.
+        Args:
+            value (str): A string for typing.
 
-        Argument is the same as WebElement#send_keys
+        Returns:
+            `appium.webdriver.webelement.WebElement`
         """
         self._execute(RemoteCommand.SEND_KEYS_TO_ELEMENT,
                       {'text': "".join(keys_to_typing(value)),
                        'value': keys_to_typing(value)})
+        return self
--- a/appium/webdriver/webelement.py
+++ b/appium/webdriver/webelement.py
@@ -216,7 +216,7 @@ class WebElement(AppiumWebElementSearchContext):
         Returns:
             `appium.webdriver.webelement.WebElement`
         """
-        _value = keys_to_typing(value)
+        keys = keys_to_typing(value)
         self._execute(RemoteCommand.SEND_KEYS_TO_ELEMENT,
-                      {'text': "".join(_value), 'value': _value})
+                      {'text': ''.join(keys), 'value': keys})
         return self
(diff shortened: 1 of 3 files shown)


appium/webdriver/webelement.py
```diff
@@ -204,3 +205,16 @@ def set_value(self, value):
         }
         self._execute(Command.SET_IMMEDIATE_VALUE, data)
         return self
+
+    def send_keys_direct(self, *value):
```
mykola-mokhnach, line 209: I'd rather overload the original selenium's implementation of send_keys. We don't support sending files in neither of Appium drivers.

KazuCocoa: hm, for mobile browser, too? I considered browser cases. (I know we don't support it for native apps)

I found Ruby binding handles the behaviour with a flag, so I wondered if we also do a similar thing for send_keys if mobile safari/chrome can work as same as desktop browser.

mykola-mokhnach: Safari mobile driver for sure does not support local files upload. Let me check chromedriver sources

mykola-mokhnach: Chromedriver handles the path inside sendKeys handler: https://github.com/bayandin/chromedriver/blob/47f9916496782bf8c1a031a3c1f1f0c7cf654da1/element_commands.cc#L452

This means there is also no need to modify/patch the call on the client side

mykola-mokhnach: We've never declared file upload support anyway: https://github.com/appium/appium/issues/13156

However, in case of mobile Chromedriver (at least according to what is present in its code) it might be possible to provide a remote file path into `sendKeys` endpoint and in case the target input is of type `file` and the remote path exists (the file must be uploaded to the device and located somewhere, where it would accessible for the browser or webview owner process) the feature should work

KazuCocoa: Got it 👍 
Thanks for the investigation

test/functional/ios/keyboard_tests.py
```diff
@@ -52,7 +52,7 @@ def test_hide_keyboard_no_key_name(self):
 
         el = self.driver.find_elements_by_class_name('XCUIElementTypeTextField')[0]
         el.click()
-        el.set_value('Testing')
+        el.send_keys('Testing')
```
KazuCocoa, line 55: Use `send_keys` randomly as part of functional test

mykola-mokhnach: it makes sense to provide a path to an existing temporary file to make sure we handle these properly

appium/webdriver/webelement.py
```diff
@@ -204,3 +205,18 @@ def set_value(self, value):
...
+        """Simulates typing into the element.
+
+        Args:
+            value (str): A string for typing.
+
+        Returns:
+            `appium.webdriver.webelement.WebElement`
+        """
+        self._execute(RemoteCommand.SEND_KEYS_TO_ELEMENT,
+                      {'text': "".join(keys_to_typing(value)),
```
mykola-mokhnach, line 220: `keys_to_typing(value)` could be precalculated to avoid duplication

appium/webdriver/webelement.py
```diff
@@ -216,7 +216,7 @@ def send_keys(self, *value):
         Returns:
             `appium.webdriver.webelement.WebElement`
         """
+        _value = keys_to_typing(value)
```
mykola-mokhnach, line 219: how about `_value -> keys`?

PR #531, opened 2020-05-05
test: Add appium_service functional test

ki4070ma:
### Details

Added appium_service functional test


test/functional/android/appium_service_tests.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
```
mykola-mokhnach, line 2: this is not needed anymore for python3

ki4070ma: I'll check and remove this line if possible. (not only here)

PR #536, opened 2020-05-08
fix: Tune mixin types, so linters could recognize them better

mykola-mokhnach:
This should improve auto-complete suggestions in the IDE and address type warnings in PyCharm.

It would be necessary to use [protocols](https://www.python.org/dev/peps/pep-0544/) if we want to give mypy a hint about access to other mixin classes. For example:

```
class CanValidateCoordinates(Protocol):
    def validate(x, y): ...

......

Here we have some other mixin class, which implements the `CanValidateCoordinates` protocol
......
T = TypeVar('T', bound=Union['MixinClass', CanValidateCoordinates])
....
def mixin_method(self: T):
   self.validate(x, y)  # <--- mypy appreciates

```

### DIFF
--- a/appium/webdriver/extensions/screen_record.py
+++ b/appium/webdriver/extensions/screen_record.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Any, Union, TypeVar, TYPE_CHECKING
+from typing import TYPE_CHECKING, Any, TypeVar, Union
 
 from selenium import webdriver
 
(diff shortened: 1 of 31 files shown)


appium/webdriver/extensions/screen_record.py
```diff
@@ -12,16 +12,22 @@
...
 
-from typing import Any, Union
+from typing import TYPE_CHECKING, Any, TypeVar, Union
 
 from selenium import webdriver
 
 from ..mobilecommand import MobileCommand as Command
 
+if TYPE_CHECKING:
+    # noinspection PyUnresolvedReferences
```
ki4070ma, line 22: [Question] Is this for Pycharm?

mykola-mokhnach: exactly. Pycharm has its own linter, which shows warnings

PR #545, opened 2020-05-18
fix: broken package

KazuCocoa:
Manually installed by `python setup.py install` on local package, then all files were in `build/lib/appium/**`

image

We can check the package with:

1. `python setup.py sdist`
2. Go to `dist/<package>`
3. `python setup.py install`
4. Check the number of files in ``build/lib/appium/**`

Let me add the package check in release script.


KazuCocoa:
Current master has only a few directories in `build/lib/appium/**`

Issue #542, opened 2020-05-17
module 'appium.webdriver' has no attribute 'Remote'


refaelsh:
## The problem

I have the same issue as this: https://github.com/appium/python-client/issues/476.
His solution is not applicable to my case.

Here is the shortest possible code that reproduces the issue:

```
from appium import webdriver

driver = webdriver.Remote('http://localhost:4723/wd/hub', [])
```

The result is like this:

```
C:\Python36\python.exe C:/Users/rsheink/PycharmProjects/untitled/bla.py
Traceback (most recent call last):
  File "C:/Users/rsheink/PycharmProjects/untitled/bla.py", line 3, in <module>
    driver = webdriver.Remote('http://localhost:4723/wd/hub', [])
AttributeError: module 'appium.webdriver' has no attribute 'Remote'

Process finished with exit code 1
```

Please help :-)

GlennJC:
@refaelsh Can you confirm what version of the Appium Python client you are using? 

Did a new install this morning and got v1.0.0. From what I can see so far, there is something broken in this release. Code that has been working for months completely stopped. There is no webdriver.Remote (same issue you are having), and the common namespace has completely disappeared (I stopped looking at that point). Went back to v0.52 and everything is working again.

I'm using Python 3.8.2

Aderlx:
what's wrong with v1.00?

GlennJC:
> 
> 
> what's wrong with v1.00?

Its in my message, however consider the following code that passes validation under 0.52:

``` python
from appium import webdriver
from appium.webdriver.common.mobileby import MobileBy

desired_caps = {"app": "Microsoft.WindowsCalculator_8wekyb3d8bbwe!App"}
driver = webdriver.Remote(
    command_executor='http://127.0.0.1:4723',
    desired_capabilities=desired_caps)

control = (MobileBy.ACCESSIBILITY_ID, 'MenuItemsHost')
```

When run under 0.52, works fine.  Under v1.0.0..

```
C:\Testing>python appium_test.py
Traceback (most recent call last):
  File "appium_test.py", line 2, in <module>
    from appium.webdriver.common.mobileby import MobileBy
ModuleNotFoundError: No module named 'appium.webdriver.common'

Traceback (most recent call last):
  File "appium_test.py", line 5, in <module>
    driver = webdriver.Remote(
AttributeError: module 'appium.webdriver' has no attribute 'Remote'
```

[...]

PR #547, opened 2020-05-18
chore: add checking package file count comparison in release script

KazuCocoa:
It help to prevent broken package like https://github.com/appium/python-client/issues/542 without any expensive process.

### DIFF
--- a/script/release.py
+++ b/script/release.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 """Release script to publish release module to pipy."""
 
+import glob
 import io
 import os
 import shutil
@@ -115,27 +116,30 @@ def build() -> None:
 
 
 def get_py_files_in_dir(root_dir: str) -> List[str]:
-    _files = []
-    for (dir_path, _dirs, files) in os.walk(root_dir):
-        for file_name in files:
-            _, file_extension = os.path.splitext(file_name)
-            if file_extension in ['.py', '.typed']:
-                _files.append(os.path.join(dir_path, file_name))
-    base_dir_path_count = len(root_dir)
-    return [_file[base_dir_path_count:] for _file in _files]
+    return [
+        file_path[len(root_dir):]
+        for file_path in glob.glob(f"{root_dir}/**/*.py", recursive=True) + glob.glob(f"{root_dir}/**/*.typed", recursive=True)
+    ]
 
 
-def compare_file_count_in_package() -> None:
+def assert_files_count_in_package() -> None:
     original_files = get_py_files_in_dir(APPIUM_DIR_PATH)
     built_files = get_py_files_in_dir(BUILT_APPIUM_DIR_PATH)
+
+    print(len(original_files))
+    print(len(built_files))
+
     if len(original_files) != len(built_files):
         print(f"The count of files in '{APPIUM_DIR_PATH}' and '{BUILT_APPIUM_DIR_PATH}' were different.")
 
-        diff = set(original_files).difference(set(built_files))
-        if diff != set():
+        original_files_set = set(original_files)
+        built_files_set = set(built_files)
+
+        diff = original_files_set.difference(built_files_set)
+        if diff:
             print(f"'{APPIUM_DIR_PATH}' has '{diff}' files than {BUILT_APPIUM_DIR_PATH}")
-        diff = set(built_files).difference(set(original_files))
-        if diff != set():
+        diff = built_files_set.difference(original_files_set)
+        if diff:
             print(f"{BUILT_APPIUM_DIR_PATH} has {diff} files than {APPIUM_DIR_PATH}")
 
         exit(f"Python files in '{BUILT_APPIUM_DIR_PATH}' may differ from '{APPIUM_DIR_PATH}'. "
@@ -151,7 +155,7 @@ def main():
     update_version_file(new_version)
 
     build()
-    compare_file_count_in_package()
+    assert_files_count_in_package()
 
     ensure_publication(new_version)
 
--- a/script/release.py
+++ b/script/release.py
@@ -15,10 +15,14 @@
 
 import io
 import os
+import shutil
 import sys
 
 VERSION_FILE_PATH = os.path.join(os.path.dirname('__file__'), 'appium', 'version.py')
 CHANGELOG_PATH = os.path.join(os.path.dirname('__file__'), 'CHANGELOG.rst')
+
+APPIUM_DIR_PATH = os.path.join(os.path.dirname('__file__'), 'appium')
+BUILT_APPIUM_DIR_PATH = os.path.join(os.path.dirname('__file__'), 'build', 'lib', 'appium')
 
 MESSAGE_RED = '\033[1;31m{}\033[0m'
 MESSAGE_GREEN = '\033[1;32m{}\033[0m'
@@ -101,6 +105,26 @@ def validate_release_env():
         exit("Please get twine via 'pip install gitchangelog' or 'pip install git+git://github.com/vaab/gitchangelog.git' for Python 3.7")
 
 
+def build():
+    shutil.rmtree(BUILT_APPIUM_DIR_PATH, ignore_errors=True)
+    os.system('{} setup.py install'.format(os.getenv('PYTHON_BIN_PATH')))
+
+
+def compare_file_count():
+    def _get_py_files(root_dir):
+        return_files = []
+        for (dir_path, _dirs, files) in os.walk(root_dir):
+            for file_name in files:
+                _, file_extension = os.path.splitext(file_name)
+                if file_extension in ['.py', '.typed']:
+                    return_files.append('{}/{}'.format(dir_path, file_name))
+        return return_files
+
+    original_files = _get_py_files(APPIUM_DIR_PATH)
+    built_files = _get_py_files(BUILT_APPIUM_DIR_PATH)
+    return len(original_files) - 1 == len(built_files)
+
+
 def main():
     validate_release_env()
 
@@ -108,6 +132,11 @@ def main():
     new_version = get_new_version()
 
     update_version_file(new_version)
+
+    build()
+    if not compare_file_count():
+        exit("Python files in 'build/lib/appium' directory may differ from 'appium'. "
+             "Please make sure setup.py package section is expected.")
 
     ensure_publication(new_version)
 
--- a/script/release.py
+++ b/script/release.py
@@ -16,6 +16,7 @@
 import io
 import os
 import shutil
+import subprocess
 import sys
 from typing import List
 
@@ -108,7 +109,9 @@ def validate_release_env():
 
 def build() -> None:
     shutil.rmtree(BUILT_APPIUM_DIR_PATH, ignore_errors=True)
-    os.system('{} setup.py install'.format(os.getenv('PYTHON_BIN_PATH')))
+    status, output = subprocess.getstatusoutput('{} setup.py install'.format(os.getenv('PYTHON_BIN_PATH')))
+    if status != 0:
+        exit(f'Failed to build the package:\n{output}')
 
 
 def get_py_files_in_dir(root_dir: str) -> List[str]:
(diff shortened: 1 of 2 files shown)


script/release.py
```diff
@@ -101,6 +105,31 @@ def validate_release_env():
...
+
+
+def compare_file_count():
+    def _get_py_files(root_dir):
+        return_files = []
+        for (dir_path, _dirs, files) in os.walk(root_dir):
+            for file_name in files:
+                _, file_extension = os.path.splitext(file_name)
+                if file_extension in ['.py', '.typed']:
+                    return_files.append('{}/{}'.format(dir_path, file_name))
```
mykola-mokhnach, line 120: use os.path.join

script/release.py
```diff
@@ -101,6 +105,31 @@ def validate_release_env():
...
 
 
+def build():
+    shutil.rmtree(BUILT_APPIUM_DIR_PATH, ignore_errors=True)
+    os.system('{} setup.py install'.format(os.getenv('PYTHON_BIN_PATH')))
+
+
+def compare_file_count():
+    def _get_py_files(root_dir):
+        return_files = []
```
mykola-mokhnach, line 115: matched_paths

script/release.py
```diff
@@ -101,6 +105,31 @@ def validate_release_env():
         exit("Please get twine via 'pip install gitchangelog' or 'pip install git+git://github.com/vaab/gitchangelog.git' for Python 3.7")
 
 
+def build():
+    shutil.rmtree(BUILT_APPIUM_DIR_PATH, ignore_errors=True)
+    os.system('{} setup.py install'.format(os.getenv('PYTHON_BIN_PATH')))
+
+
+def compare_file_count():
+    def _get_py_files(root_dir):
```
mykola-mokhnach, line 114: I'd rather move this function to the top module level

script/release.py
```diff
@@ -101,6 +105,31 @@ def validate_release_env():
...
+        return_files = []
+        for (dir_path, _dirs, files) in os.walk(root_dir):
+            for file_name in files:
+                _, file_extension = os.path.splitext(file_name)
+                if file_extension in ['.py', '.typed']:
+                    return_files.append('{}/{}'.format(dir_path, file_name))
+        return return_files
+
+    original_files = _get_py_files(APPIUM_DIR_PATH)
+    built_files = _get_py_files(BUILT_APPIUM_DIR_PATH)
```
mykola-mokhnach, line 124: we probably don't have to return file names, paths, but only counts

script/release.py
```diff
@@ -101,6 +105,31 @@ def validate_release_env():
         exit("Please get twine via 'pip install gitchangelog' or 'pip install git+git://github.com/vaab/gitchangelog.git' for Python 3.7")
 
 
+def build():
+    shutil.rmtree(BUILT_APPIUM_DIR_PATH, ignore_errors=True)
+    os.system('{} setup.py install'.format(os.getenv('PYTHON_BIN_PATH')))
+
+
+def compare_file_count():
```
mykola-mokhnach, line 113: consider using typing

script/release.py
```diff
@@ -101,6 +105,31 @@ def validate_release_env():
...
+        for (dir_path, _dirs, files) in os.walk(root_dir):
+            for file_name in files:
+                _, file_extension = os.path.splitext(file_name)
+                if file_extension in ['.py', '.typed']:
+                    return_files.append('{}/{}'.format(dir_path, file_name))
+        return return_files
+
+    original_files = _get_py_files(APPIUM_DIR_PATH)
+    built_files = _get_py_files(BUILT_APPIUM_DIR_PATH)
+    if len(original_files) != len(built_files):
```
mykola-mokhnach, line 125: I'd rather check that original files count <= built files count

script/release.py
```diff
@@ -101,6 +105,31 @@ def validate_release_env():
...
+        print("The count of files in 'build/lib/appium' and 'appium' were different. "
+              "'appium': {},\n 'build/lib/appium': {}".format(original_files, built_files))
```
mykola-mokhnach, line 127: consider using python format strings

script/release.py
mykola-mokhnach: this function is expected to throw an exception in case of failure. Consider renaming it to `is_...` if it is supposed to return boolean

script/release.py
```diff
@@ -109,6 +138,11 @@ def main():
 
     update_version_file(new_version)
 
+    build()
+    if not compare_file_count():
+        exit("Python files in 'build/lib/appium' directory may differ from 'appium'. "
+             "Please make sure setup.py package section is expected.")
```
mykola-mokhnach, line 144: Please make sure `setup.py` is configured properly

script/release.py
KazuCocoa: I considered which was better, but wanted to check the list when some packages was missed. The listed error message helped me to see the diff in #542 case. No strong opinion, but I won't write the similar code to get files when error happens for now. (Current count is not so huge like over 100)

script/release.py
KazuCocoa: yeah, I also thought so, but for now, the count should be the same.
We can relax/tweak it when we get a situation they becomes different

script/release.py
mykola-mokhnach: makes sense. In such case I'd consider returning sets instead and only show the actual difference between these sets (using `difference` helper)

script/release.py
```diff
@@ -101,6 +106,32 @@ def validate_release_env():
         exit("Please get twine via 'pip install gitchangelog' or 'pip install git+git://github.com/vaab/gitchangelog.git' for Python 3.7")
 
 
+def build() -> None:
+    shutil.rmtree(BUILT_APPIUM_DIR_PATH, ignore_errors=True)
+    os.system('{} setup.py install'.format(os.getenv('PYTHON_BIN_PATH')))
```
mykola-mokhnach, line 111: consider used subprocess module to start processes

script/release.py
```diff
@@ -101,6 +106,32 @@ def validate_release_env():
...
+
+
+def _get_py_files(root_dir: str) -> List[str]:
```
mykola-mokhnach, line 114: underscores are not necessary

script/release.py
```diff
@@ -101,6 +106,32 @@ def validate_release_env():
...
+            if file_extension in ['.py', '.typed']:
+                _files.append(os.path.join(dir_path, file_name))
+    return _files
+
+
+def has_same_file_count_original_built() -> bool:
+    original_files = _get_py_files(APPIUM_DIR_PATH)
+    built_files = _get_py_files(BUILT_APPIUM_DIR_PATH)
+    if len(original_files) != len(built_files):
+        print("The count of files in 'build/lib/appium' and 'appium' were different. "
```
mykola-mokhnach, line 128: I'd rather put variables here instead of hardcoded folder paths

script/release.py
```diff
@@ -109,6 +140,11 @@ def main():
 
     update_version_file(new_version)
 
+    build()
+    if not has_same_file_count_original_built():
+        exit("Python files in 'build/lib/appium' directory may differ from 'appium'. "
```
mykola-mokhnach, line 145: ⬆️

script/release.py
KazuCocoa: Good idea

script/release.py
```diff
@@ -108,28 +109,37 @@ def validate_release_env():
...
     shutil.rmtree(BUILT_APPIUM_DIR_PATH, ignore_errors=True)
-    os.system('{} setup.py install'.format(os.getenv('PYTHON_BIN_PATH')))
+    status, output = subprocess.getstatusoutput('{} setup.py install'.format(os.getenv('PYTHON_BIN_PATH')))
+    if status != 0:
+        exit(f'Failed to build the package:\n{output}')
 
 
-def _get_py_files(root_dir: str) -> List[str]:
+def get_py_files_in_dir(root_dir: str) -> List[str]:
     _files = []
```
mykola-mokhnach, line 118: please remove all these starting underscores

script/release.py
```diff
@@ -108,28 +109,37 @@ def validate_release_env():
...
+    built_files = get_py_files_in_dir(BUILT_APPIUM_DIR_PATH)
     if len(original_files) != len(built_files):
-        print("The count of files in 'build/lib/appium' and 'appium' were different. "
-              f"'appium': {original_files},\n 'build/lib/appium': {built_files}")
-        return False
+        print(f"The count of files in '{APPIUM_DIR_PATH}' and '{BUILT_APPIUM_DIR_PATH}' were different.")
 
-    return True
+        diff = set(original_files).difference(set(built_files))
+        if diff != set():
```
mykola-mokhnach, line 135: simply `if diff`. Empty set is considered falsy

script/release.py
```diff
@@ -108,28 +109,37 @@ def validate_release_env():
...
+            print(f"'{APPIUM_DIR_PATH}' has '{diff}' files than {BUILT_APPIUM_DIR_PATH}")
+        diff = set(built_files).difference(set(original_files))
+        if diff != set():
```
mykola-mokhnach, line 138: ⬆️

script/release.py
```diff
@@ -108,28 +109,37 @@ def validate_release_env():
...
-        print("The count of files in 'build/lib/appium' and 'appium' were different. "
-              f"'appium': {original_files},\n 'build/lib/appium': {built_files}")
-        return False
+        print(f"The count of files in '{APPIUM_DIR_PATH}' and '{BUILT_APPIUM_DIR_PATH}' were different.")
 
-    return True
+        diff = set(original_files).difference(set(built_files))
+        if diff != set():
+            print(f"'{APPIUM_DIR_PATH}' has '{diff}' files than {BUILT_APPIUM_DIR_PATH}")
+        diff = set(built_files).difference(set(original_files))
```
mykola-mokhnach, line 137: both sets could be precalculated

script/release.py
```diff
@@ -141,9 +151,7 @@ def main():
     update_version_file(new_version)
 
     build()
-    if not has_same_file_count_original_built():
-        exit("Python files in 'build/lib/appium' directory may differ from 'appium'. "
-             "Please make sure setup.py is configured properly.")
+    compare_file_count_in_package()
```
mykola-mokhnach, line 154: 👍

mykola-mokhnach: I'd say assert_files_count_in_package

script/release.py
```diff
@@ -101,6 +107,41 @@ def validate_release_env():
...
 
 
+def build() -> None:
+    shutil.rmtree(BUILT_APPIUM_DIR_PATH, ignore_errors=True)
+    status, output = subprocess.getstatusoutput('{} setup.py install'.format(os.getenv('PYTHON_BIN_PATH')))
+    if status != 0:
+        exit(f'Failed to build the package:\n{output}')
+
+
+def get_py_files_in_dir(root_dir: str) -> List[str]:
```
ki4070ma, line 117: You can do the same thing as below. (maybe same?)

```python
import glob
import os
files = glob.glob(f'{root_dir}/**/*.py', recursive=True) + glob.glob(f'{root_dir}/**/*.typed', recursive=True)
return [os.path.basename(x) for x in files]
```

KazuCocoa: ah, recursive is available in 3

script/release.py
```diff
@@ -115,27 +116,30 @@ def build() -> None:
...
+        for file_path in glob.glob(f"{root_dir}/**/*.py", recursive=True) + glob.glob(f"{root_dir}/**/*.typed", recursive=True)
+    ]
 
 
-def compare_file_count_in_package() -> None:
+def assert_files_count_in_package() -> None:
     original_files = get_py_files_in_dir(APPIUM_DIR_PATH)
     built_files = get_py_files_in_dir(BUILT_APPIUM_DIR_PATH)
+
+    print(len(original_files))
```
mykola-mokhnach, line 129: Please also mention what is printed: `f'Original files count: {len(original_files)}'`

script/release.py
```diff
@@ -115,27 +116,30 @@ def build() -> None:
...
+
     if len(original_files) != len(built_files):
         print(f"The count of files in '{APPIUM_DIR_PATH}' and '{BUILT_APPIUM_DIR_PATH}' were different.")
 
-        diff = set(original_files).difference(set(built_files))
-        if diff != set():
+        original_files_set = set(original_files)
+        built_files_set = set(built_files)
+
+        diff = original_files_set.difference(built_files_set)
```
mykola-mokhnach, line 138: you could also use `-`:  https://stackoverflow.com/questions/30986751/difference-between-set-difference-and-set-subtraction

script/release.py
KazuCocoa: oops, this is my debugging...

script/release.py
```diff
@@ -115,27 +116,30 @@ def build() -> None:
...
-    _files = []
-    for (dir_path, _dirs, files) in os.walk(root_dir):
-        for file_name in files:
-            _, file_extension = os.path.splitext(file_name)
-            if file_extension in ['.py', '.typed']:
-                _files.append(os.path.join(dir_path, file_name))
-    base_dir_path_count = len(root_dir)
-    return [_file[base_dir_path_count:] for _file in _files]
+    return [
+        file_path[len(root_dir):]
```
mykola-mokhnach, line 120: so pythonic 🙀

script/release.py
KazuCocoa: ah, often used the way in Ruby. Cool

Issue #512, opened 2020-04-09
'TouchAction' object has no attribute 'flick_element'


KageyamaRuka:
## The problem

'TouchAction' object has no attribute 'flick_element'

## Environment
- Appium version (or git revision) that exhibits the issue: 1.17.0
- Desktop OS/version used to run Appium: MacOS 10.15.3
- Node.js version (unless using Appium.app|exe): v13.11.0
- Mobile platform/version under test: iOS 13.3.1
- Real device or emulator/simulator: Real device
- Appium CLI or Appium.app|exe: Appium CLI

## Link to Appium Logs

[log.txt](https://github.com/appium/python-client/files/4455646/log.txt)

## Code To reproduce issue

![image](https://user-images.githubusercontent.com/24635872/78886353-a1674380-7a90-11ea-9076-41549b2f7972.png)

ki4070ma:
Hi

You can use `flick` from `ActionHelpers`.

https://github.com/appium/python-client/blob/master/appium/webdriver/extensions/action_helpers.py#L130

Need to update docs.

http://appium.io/docs/en/commands/interactions/touch/flick/

KazuCocoa:
TouchAction is deprecated. Please use W3C actions instead.

PR #559, opened 2020-08-15
feat: Add flick_uia2

ki4070ma:
### Background
For https://github.com/appium/python-client/issues/512

### TODO
- [x] unit test
- [x] func test

### DIFF
--- a/appium/webdriver/webdriver.py
+++ b/appium/webdriver/webdriver.py
@@ -345,6 +345,50 @@ class WebDriver(
         self.execute(Command.SET_IMMEDIATE_VALUE, data)
         return self
 
+    def flick_element(self, element: Optional[MobileWebElement] = None,
+                      xspeed: int = 0, yspeed: int = 0,
+                      xoffset: int = 0, yoffset: int = 0,
+                      speed: int = 0) -> T:
+        """Flick the element
+
+        Args:
+            element: the element to be flicked
+            xoffset: used for the flick endpoint (endpoint: element x pos + xoffset)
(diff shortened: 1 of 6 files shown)


appium/webdriver/webdriver.py
```diff
@@ -345,6 +345,50 @@ def set_value(self, element: MobileWebElement, value: str) -> T:
...
         return self
 
+    def flick_element(self, element: Optional[MobileWebElement] = None,
+                      xspeed: int = 0, yspeed: int = 0,
+                      xoffset: int = 0, yoffset: int = 0,
+                      speed: int = 0) -> T:
+        """Flick the element
+
+        Args:
+            element: the element to be flicked
```
ki4070ma, line 355: [Notes] https://github.com/appium/appium-uiautomator2-server/blob/c0d2d45740a2f9d05cf0e6c3b7b1a19e2f85d1b1/app/src/main/java/io/appium/uiautomator2/handler/Flick.java#L52

mykola-mokhnach: these shortcuts were deliberately removed from java client as they are not cross-platform. 
I would rather prefer splitting platform-specific actions and assigning them to their appropriate driver classes

ki4070ma: thanks, I haven't taken care of `not cross-platform` (uia2).

KazuCocoa:
Selenium Python binding has the command, btw.
https://github.com/SeleniumHQ/selenium/blob/941dc9c6b2e2aa4f701c1b72be8de03d4b7e996a/py/selenium/webdriver/common/touch_actions.py#L168
You can simply add the same method in this touch action module

PR #565, opened 2020-12-02
ci: remove travis

KazuCocoa:
I should check tox on Azure more...
https://github.com/appium/python-client/pull/565/commits/856a82f5c581e7ec832fe66cb3272cda07a63e15 this `resources` did not work.
It raised an invalid repository error, so potentially we should copy them


azure-pipelines.yml
```diff
@@ -5,6 +5,24 @@
 jobs:
   - template: ./ci-jobs/functional_test.yml
 
+  - job: 'unitTests'
+    pool:
+      vmImage: 'ubuntu-16.04'
```
mykola-mokhnach, line 10: why so old image?

KazuCocoa: To find the cause of failing python versions... just tweaking config

PR #567, opened 2020-12-20
feat: add warning to drop forceMjsonwp for W3C

KazuCocoa:
`forceMjsonwp` only sends MJSONWP formatted capabilities, but selenium 4 and Appium is already going to focus on W3C.

The initial motivation to allow forceMjsonwp was to help users so that they were able to move to W3C from MJSONWP in their timeline. chrome driver also already defaults to W3C spec, so it's a good time to drop it.

### DIFF
--- a/appium/webdriver/webdriver.py
+++ b/appium/webdriver/webdriver.py
@@ -241,7 +241,8 @@ class WebDriver(
         """Manage capabilities whether W3C format or MJSONWP format
         """
         if _FORCE_MJSONWP in capabilities:
-            logger.warning("[Deprecated] 'forceMjsonwp' capability will be dropped")
+            logger.warning("[Deprecated] 'forceMjsonwp' capability will be dropped after switching base selenium client from v3 to v4 "\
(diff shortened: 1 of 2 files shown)


mykola-mokhnach:
I would rather have this commit for the next major version of the client. Removing this cap would enforce us to change it anyway

KazuCocoa:
Okay, I have no objection. I changed to print a warning instead as deprecated.
(The selenium-4 branch already does not have this)

appium/webdriver/webdriver.py
```diff
@@ -241,6 +241,7 @@ def _merge_capabilities(self, capabilities: Dict) -> Dict[str, Any]:
         """Manage capabilities whether W3C format or MJSONWP format
         """
         if _FORCE_MJSONWP in capabilities:
+            logger.warning("[Deprecated] 'forceMjsonwp' capability will be dropped")
```
mykola-mokhnach, line 244: I would rather add a more detailed explanation of why the capability is going to be dropped and what is necessary to do in order to be ready for it

mykola-mokhnach (approved):

PR #575, opened 2021-01-31
chore: Fix iOS app management functional tests

ki4070ma:
### Background

* It's failing after upgrading to iOS 14.3.

### DIFF
--- a/test/functional/android/activities_tests.py
+++ b/test/functional/android/activities_tests.py
@@ -32,8 +32,8 @@ class TestActivities(BaseTestCase):
         self.driver.start_activity(APIDEMO_PKG_NAME, ".ApiDemos")
         self._assert_activity_contains('Demos')
 
-        self.driver.start_activity("com.android.calculator2", ".Calculator")
-        self._assert_activity_contains('Calculator')
+        self.driver.start_activity("com.google.android.deskclock", "com.android.deskclock.DeskClock")
+        self._assert_activity_contains('Clock')
 
     def _assert_activity_contains(self, activity: str) -> None:
         current = self.driver.current_activity
--- a/test/functional/android/applications_tests.py
+++ b/test/functional/android/applications_tests.py
@@ -55,7 +55,6 @@ class TestApplications(BaseTestCase):
         app_id = self.driver.current_package
         assert self.driver.query_app_state(app_id) == ApplicationState.RUNNING_IN_FOREGROUND
         self.driver.background_app(-1)
-        sleep(1)  # to wait the app has gone
         assert self.driver.query_app_state(app_id) < ApplicationState.RUNNING_IN_FOREGROUND
         self.driver.activate_app(app_id)
         assert self.driver.query_app_state(app_id) == ApplicationState.RUNNING_IN_FOREGROUND
--- a/test/functional/android/applications_tests.py
+++ b/test/functional/android/applications_tests.py
@@ -55,6 +55,7 @@ class TestApplications(BaseTestCase):
         app_id = self.driver.current_package
         assert self.driver.query_app_state(app_id) == ApplicationState.RUNNING_IN_FOREGROUND
         self.driver.background_app(-1)
+        sleep(1)  # to wait the app has gone
         assert self.driver.query_app_state(app_id) < ApplicationState.RUNNING_IN_FOREGROUND
         self.driver.activate_app(app_id)
         assert self.driver.query_app_state(app_id) == ApplicationState.RUNNING_IN_FOREGROUND
--- a/test/functional/ios/applications_tests.py
+++ b/test/functional/ios/applications_tests.py
@@ -14,7 +14,7 @@
 
 from appium.webdriver.applicationstate import ApplicationState
 from test.functional.ios.helper.test_helper import BaseTestCase
-from test.functional.test_helper import wait_for
+from test.functional.test_helper import wait_for_condition
 
 from .helper import desired_capabilities
 
@@ -26,7 +26,7 @@ class TestWebDriver(BaseTestCase):
             return
         assert self.driver.query_app_state(desired_capabilities.BUNDLE_ID) == ApplicationState.RUNNING_IN_FOREGROUND
         self.driver.background_app(-1)
-        assert wait_for(
+        assert wait_for_condition(
             lambda: self.driver.query_app_state(desired_capabilities.BUNDLE_ID) < ApplicationState.RUNNING_IN_FOREGROUND
         ), "The app didn't go to background."
         self.driver.activate_app(desired_capabilities.BUNDLE_ID)
--- a/test/functional/ios/applications_tests.py
+++ b/test/functional/ios/applications_tests.py
@@ -27,7 +27,12 @@ class TestWebDriver(BaseTestCase):
             return
         assert self.driver.query_app_state(desired_capabilities.BUNDLE_ID) == ApplicationState.RUNNING_IN_FOREGROUND
         self.driver.background_app(-1)
-        sleep(1)  # to wait the app has gone
-        assert self.driver.query_app_state(desired_capabilities.BUNDLE_ID) < ApplicationState.RUNNING_IN_FOREGROUND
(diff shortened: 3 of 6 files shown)


test/functional/android/applications_tests.py
```diff
@@ -55,6 +55,7 @@ def test_app_management(self) -> None:
         app_id = self.driver.current_package
         assert self.driver.query_app_state(app_id) == ApplicationState.RUNNING_IN_FOREGROUND
         self.driver.background_app(-1)
+        sleep(1)  # to wait the app has gone
```
ki4070ma, line 58: Here comes from https://github.com/appium/ruby_lib_core/blame/e13c58fe4355dcb2631f0a42248d991165f428f7/test/functional/android/android/device_test.rb#L87.

(but I'll try to avoid `sleep`.)

ki4070ma: Here need to wait 1 sec even after the element is invisible.

test/functional/ios/applications_tests.py
```diff
@@ -25,6 +27,7 @@ def test_app_management(self) -> None:
             return
         assert self.driver.query_app_state(desired_capabilities.BUNDLE_ID) == ApplicationState.RUNNING_IN_FOREGROUND
         self.driver.background_app(-1)
+        sleep(1)  # to wait the app has gone
         assert self.driver.query_app_state(desired_capabilities.BUNDLE_ID) < ApplicationState.RUNNING_IN_FOREGROUND
```
mykola-mokhnach, line 31: I would rather add a timer to the background check

ki4070ma:
Will handle failed android tests with SDK 30 by https://github.com/appium/python-client/issues/576 .

test/functional/android/activities_tests.py
```diff
@@ -32,8 +32,8 @@ def test_start_activity_other_app(self) -> None:
         self.driver.start_activity(APIDEMO_PKG_NAME, ".ApiDemos")
         self._assert_activity_contains('Demos')
 
-        self.driver.start_activity("com.android.calculator2", ".Calculator")
```
ki4070ma, line 35: (note) Here is one of https://github.com/appium/python-client/issues/576.

test/functional/test_helper.py
```diff
@@ -37,3 +39,23 @@ def is_ci() -> bool:
...
+        bool: True if `method` returns True
+
+    """
+    for i in range(timeout + 1):
+        if method():
+            break
+        if i == timeout:
+            return False
+        sleep(1)
+    return True
```
ki4070ma, line 61:
```suggestion
    for i in range(timeout + 1):
        if method():
            return True
        sleep(1)
    return False
```

test/functional/test_helper.py
```diff
@@ -37,3 +39,21 @@ def is_ci() -> bool:
...
+        method: The target method to be waited. Need to return `bool`
+        timeout: The timeout to be waited (sec.)
+
+    Returns:
+        bool: True if `method` returns True
+
+    """
+    for i in range(timeout + 1):
+        if method():
+            return True
```
mykola-mokhnach, line 57: I would rather return the result of the method itself. this should not necessarily be a boolean value

test/functional/test_helper.py
```diff
@@ -37,3 +39,21 @@ def is_ci() -> bool:
         `True` if current executions is on CI
     """
     return os.getenv('CI', 'false') == 'true'
+
+
+def wait_for(method: Callable, timeout: int) -> bool:
```
mykola-mokhnach, line 44: I usually prefer to explicitly mention the time unit for variables (e.g. timeout -> timeout_sec). Also, it makes sense to have a default value (e.g. 5) for it and make it float, so we could provide timeouts lesser than one seconds. It also makes sense to make the interval configurable

test/functional/ios/applications_tests.py
```diff
@@ -12,8 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from time import sleep
```
mykola-mokhnach, line 15: is the import still used?

test/functional/ios/applications_tests.py
```diff
@@ -12,8 +12,11 @@
...
+
 from appium.webdriver.applicationstate import ApplicationState
 from test.functional.ios.helper.test_helper import BaseTestCase
+from test.functional.test_helper import wait_for
```
mykola-mokhnach, line 19: how about `wait_for_condition` ?

test/functional/test_helper.py
ki4070ma: @mykola-mokhnach 

Humm, in my understanding, I'm not sure how it can wait for target `Callable` if it isn't boolean.
Would you give more details?

test/functional/test_helper.py
mykola-mokhnach: sure:

```
result = callable()
if result:
    return result
```

mykola-mokhnach: we could even use 

```
if result := callable():
    return result
```

if we drop the support of py37 :-P

ki4070ma: @mykola-mokhnach 
Thanks for details.
Ah, yes, actually it popped into my head,
but as my thoughts, it could extend this method usage, but extended usage is limited for "user".

But it's ok since "user" is developer. I'll change it to suggestion. 👍

test/functional/test_helper.py
```diff
@@ -37,3 +39,27 @@ def is_ci() -> bool:
...
+        interval: The interval for wait (sec.)
+
+    Returns:
+        Any: value which `method` returns
+
+    """
+    result = method()
+    for i in range(int(timeout_sec / interval)):
+        if i != 0:
+            result = method
```
ki4070ma, line 61:
```suggestion
            result = method()
```

test/functional/test_helper.py
```diff
@@ -37,3 +39,27 @@ def is_ci() -> bool:
...
+        timeout: The timeout to be waited (sec.)
+        interval: The interval for wait (sec.)
+
+    Returns:
+        Any: value which `method` returns
+
+    """
+    result = method()
+    for i in range(int(timeout_sec / interval)):
+        if i != 0:
```
ki4070ma, line 60: 🤔

test/functional/test_helper.py
```diff
@@ -37,3 +39,27 @@ def is_ci() -> bool:
...
+        method: The target method to be waited
+        timeout: The timeout to be waited (sec.)
+        interval: The interval for wait (sec.)
+
+    Returns:
+        Any: value which `method` returns
+
+    """
+    result = method()
+    for i in range(int(timeout_sec / interval)):
```
mykola-mokhnach, line 59: I would rather use while loop and compare with the time.time() value

test/functional/test_helper.py
```diff
@@ -54,11 +55,16 @@ def wait_for_condition(method: Callable, timeout_sec: float = 5, interval: float
...
 
+    Raises:
+        ValueError: When interval isn't more than 0
+
     """
-    result = method()
-    for i in range(int(timeout_sec / interval)):
-        if i != 0:
-            result = method()
+    if interval <= 0:
```
mykola-mokhnach, line 62: 👍

test/functional/test_helper.py
```diff
@@ -54,11 +55,16 @@ def wait_for_condition(method: Callable, timeout_sec: float = 5, interval: float
...
+        raise ValueError('interval needs to be more than 0')
+
+    started = time.time()
+    while time.time() - started <= timeout_sec:
+        result = method()
         if result:
             break
         sleep(interval)
```
mykola-mokhnach, line 70: interval_sec

mykola-mokhnach (approved):

test/functional/test_helper.py
mykola-mokhnach: zero interval is also fine

KazuCocoa (approved):

PR #577, opened 2021-02-06
chore: Update pipfile to respect isort v5

ki4070ma:
### Background

Since isort v4 is used.

### Note

Currently mypy target is only `functional` under `test.`( `test/unit` is ignored by mypy v0.782 somehow.)
(But in my opinion, it isn't so impact even if `test/unit` is not target.)

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -4,25 +4,25 @@ url = "https://pypi.org/simple"
 verify_ssl = true
 
 [dev-packages]
-pre-commit = "~=2.6"
+pre-commit = "~=2.10"
 
 [packages]
 selenium = "~=3.141"
 
 black = "==20.8b1"
 
-pytest = "~=6.0"
-pytest-cov = "~=2.10"
+pytest = "~=6.2"
+pytest-cov = "~=2.11"
 
-tox = "~=3.19"
+tox = "~=3.21"
 tox-travis = "~=0.12"
 
 httpretty = "~=1.0"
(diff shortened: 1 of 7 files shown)


Pipfile
```diff
@@ -4,25 +4,25 @@ url = "https://pypi.org/simple"
 verify_ssl = true
 
 [dev-packages]
-pre-commit = "~=2.6"
+pre-commit = "~=2.10"
 
 [packages]
 selenium = "~=3.141"
 
 black = "==20.8b1"
```
ki4070ma, line 12: (note) `black = "~=20.8b1"` doesn't work.

Pipfile
```diff
@@ -4,25 +4,25 @@ url = "https://pypi.org/simple"
 verify_ssl = true
 
 [dev-packages]
-pre-commit = "~=2.6"
+pre-commit = "~=2.10"
```
mykola-mokhnach, line 7: could dependabot do that?

KazuCocoa (approved):
the change seems good

mykola-mokhnach (approved):

Pipfile
ki4070ma: Yes.
I don't remember the reason to stop using it now, but previously we used it.

Can we use it again? > @KazuCocoa 

https://github.com/appium/python-client/pulls?q=is%3Apr+is%3Aclosed+author%3Aapp%2Fdependabot-preview

PR #581, opened 2021-02-08
ci: remove travis2

ki4070ma:
### Background

from https://github.com/appium/python-client/pull/565

### Tests
https://dev.azure.com/AppiumCI/Appium%20CI/_build/results?buildId=13831&view=results
* unitTests

### Ref
https://docs.microsoft.com/en-us/azure/devops/pipelines/ecosystems/python?view=azure-devops#run-tests-with-tox


Makefile
```diff
@@ -24,7 +24,7 @@ mypy:  ## Run mypy
 
 .PHONY: unittest
 unittest: ## Run unittest
-	python -m pytest test/unit/
+	python -m pytest ${ARGS} test/unit/
```
mykola-mokhnach, line 27: `$ARGS` or `@{ARGS}` ?

azure-pipelines.yml
```diff
@@ -4,6 +4,32 @@
...
+       vmImage: 'ubuntu-18.04'
+    strategy:
+      matrix:
+        Python37:
+          python.version: '3.7'
+        Python38:
+          python.version: '3.8'
+        Python39:
+          python.version: '3.9'
+      maxParallel: 2
```
mykola-mokhnach, line 18: why do we limit the amount of parallel tasks?

mykola-mokhnach (approved):

Makefile
ki4070ma: at least it can be `$ARGS` (not sure `@{ARGS}`), but will use `$(ARGS)` as well as others in this script.

azure-pipelines.yml
ki4070ma: Just copy-pasted one remains, so I can remove.

azure-pipelines.yml
```diff
@@ -4,6 +4,31 @@
 # https://docs.microsoft.com/azure/devops/pipelines/languages/xcode
 jobs:
   - template: ./ci-jobs/functional_test.yml
+  - job: 'unitTests'  # TODO Decide how to handle tox
```
KazuCocoa, line 7: Or we no longer need `tox` stuff?
I did some ways in https://github.com/appium/python-client/pull/565 like checking out scripts from their repository or adding related things in this repository. But all of them failed.

ki4070ma: > Or we no longer need tox stuff?

In my opinion for now, it's unnecessary for ci, but it's useful for running py37-39 in local env.
(current tox usage doesn't match with azure-pipeline since  just `tox` includes running unit test in py37-39, as you know.)

When we look into sample for details, we might be able to use `tox` in azure pipeline as well.

https://github.com/tox-dev/azure-pipelines-template

KazuCocoa: Probably we should tweak something more to use the template.
I tried it before, but Azure had an error https://github.com/appium/python-client/pull/565/commits/856a82f5c581e7ec832fe66cb3272cda07a63e15
(I forgot the exact error message, but maybe it was not able to get the template stuff)

Then, it would be nice to address only the local environment about `tox` in README.
https://github.com/appium/python-client#run-tests

ki4070ma: ok, I'll try it while this PR is on hold.

azure-pipelines.yml
```diff
@@ -4,6 +4,35 @@
...
+        Python38:
+          python.version: '3.8'
+        Python39:
+          python.version: '3.9'
+
+    steps:
+      - task: UsePythonVersion@0
+        displayName: 'Use Python $(python.version)'
+        inputs:
+          versionSpec: '$(python.version)'
```
KazuCocoa, line 23: oh, nice.
When I did this matrix style, I got no matched python error. But it seems fixed on Azure side(?)

KazuCocoa (approved):

PR #590, opened 2021-02-23
Fix mypy error with mypy v0.812

ki4070ma:
### Background

Root cause is that `MobileSwitchTo` is unreachable when `driver.switch_to` is called from checking current behavior by mypy/on IDE.
(Also `context`@MobileSwithTo doesn't show at IDE(PyCharm) as auto completion)

### Notes

mypy errors

```
************* Module test.functional.android.context_switching_tests
test/functional/android/context_switching_tests.py:49:8: E1101: Instance of 'SwitchTo' has no 'context' member (no-member)
test/functional/android/context_switching_tests.py:54:12: E1101: Instance of 'SwitchTo' has no 'context' member (no-member)
test/functional/android/context_switching_tests.py:59:8: E1101: Instance of 'SwitchTo' has no 'context' member (no-member)
```


appium/webdriver/switch_to.py
```diff
@@ -22,7 +22,7 @@
 
 
 class MobileSwitchTo(SwitchTo):
-    def context(self, context_name: str) -> T:
+    def context(self, context_name: Optional[str]) -> T:
```
mykola-mokhnach, line 25: why context name is optional?

ki4070ma: It comes from https://github.com/appium/python-client/blob/8eb4c3f7c2cb95ee81eb9c664de265edf473d9dc/test/functional/android/context_switching_tests.py#L49

mykola-mokhnach: it makes sense to mention that in the docstring (e.g. passing None is equal to switch to native context)

appium/webdriver/webdriver.py
```diff
@@ -160,7 +160,6 @@ def __init__(
             self._addCommands()
 
         self.error_handler = MobileErrorHandler()
-        self._switch_to = MobileSwitchTo(self)
```
mykola-mokhnach, line 163: do we use `_switch_to` property anywhere else?

ki4070ma: Thanks for your pointing out, I've checked it again, but it isn't used anywhere else.

mykola-mokhnach (approved):

PR #608, opened 2021-06-05
feat: allow to add a command dynamically

KazuCocoa:
Appium 2.0 will allow custom drivers/plugins add their own endpoint.
Then, we want to allow clients to add custom commands for them.

In Ruby, I've added like https://github.com/appium/ruby_lib_core/blob/b9f015d7dea14964a0733f2385ebcff68da1e18e/test/unit/android/webdriver/w3c/commands_test.rb#L31-L98 as an experimental method. (The method was already used internally, so as Ruby client, it was just exposed the method as a public method.)

In Python, I've added two methods. `add_command` and `execute_custom_command`.

`add_command` adds the pair of HTTP method and a path in `self.command_executor._commands`.
`execute_custom_command` calls a method from `self.command_executor._commands` as `self.execute`.
We want to hide `self.execute` and `self.command_executor` since they are for internal usage so far. They may change by updating selenium library, for instance.

```python
# add commands
driver.add_command(method='POST', url='session/$sessionId/path/to/custom/url', name='test_command')

def do_command():
    # some logic for arguments
    return_value = driver.execute_custom_command('test_command', arguments)
    # dom some process for the return value
    return result

driver.custom_my_command = do_command
...
result = driver.custom_my_command()  # just call it
...
```

I considered decoration way, but so far, this exposing two methods as public methods are simpler I think.

What do you think?
Your thought is welcome such as better naming etc.

### DIFF
--- a/appium/webdriver/webdriver.py
+++ b/appium/webdriver/webdriver.py
@@ -58,6 +58,8 @@ from .mobilecommand import MobileCommand as Command
 from .switch_to import MobileSwitchTo
 from .webelement import WebElement as MobileWebElement
 
+AVAILABLE_METHOD = frozenset(['GET', 'HEAD', 'POST', 'PUT', 'DELETE', 'CONNECT', 'OPTIONS', 'TRACE', 'PATCH'])
+
 # From remote/webdriver.py
 _W3C_CAPABILITY_NAMES = frozenset(
     [
@@ -357,14 +359,42 @@ class WebDriver(
         return MobileSwitchTo(self)
 
     def add_command(self, method: str, url: str, name: str) -> Callable:
+        """Add a custom command as 'name'
+
+        Args:
+            method: The method of HTTP request. Available methods are AVAILABLE_METHOD.
+            url: The url to URL template as https://www.w3.org/TR/webdriver/#endpoints.
+                '$sessionId' is a placeholder of current session id.
+                '$id' is a placeholder of an element.
+            name: The name of command to call in `execute_custom_command`.
+
+        Returns:
+            `appium.webdriver.webdriver.WebDriver`: Self instance
+
+        """
         if name in self.command_executor._commands:
             raise ValueError("{} is already defined".format(name))
+
+        method = method.upper()
+        if method not in AVAILABLE_METHOD:
+            raise ValueError("'{}' is invalid. Valid method is {}.".format(method, AVAILABLE_METHOD))
+
         self.command_executor._commands[name] = (method, url)
         return self
 
-    def execute_custom_command(self, name: str, args: Dict) -> Any:
+    def execute_custom_command(self, name: str, args: Dict = {}) -> Any:
+        """Execute a custom command defined as 'name' with args command.
+
+        Args:
+            name: The name of command defined by `add_command`.
+            args: The argument as this command body
+
+        Returns:
+            'value' JSON object field in the response body.
+
+        """
         if name not in self.command_executor._commands:
-            raise ValueError("No {} custom command. Please add it with 'add_command'".format(name))
+            raise ValueError("No {} custom command. Please add it by 'add_command'".format(name))
         return self.execute(name, args)['value']
 
     # pylint: disable=protected-access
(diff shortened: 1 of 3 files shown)


appium/webdriver/webdriver.py
```diff
@@ -58,6 +58,8 @@
 from .switch_to import MobileSwitchTo
 from .webelement import WebElement as MobileWebElement
 
+AVAILABLE_METHOD = frozenset(['GET', 'HEAD', 'POST', 'PUT', 'DELETE', 'CONNECT', 'OPTIONS', 'TRACE', 'PATCH'])
```
mykola-mokhnach, line 61: perhaps AVAILABLE_METHODS?

appium/webdriver/webdriver.py
```diff
@@ -356,6 +358,47 @@ def switch_to(self) -> MobileSwitchTo:
 
         return MobileSwitchTo(self)
 
+    def add_command(self, method: str, url: str, name: str) -> Callable:
```
mykola-mokhnach, line 361: is self of type Callable?

appium/webdriver/webdriver.py
mykola-mokhnach: this could be an Enum

test/unit/webdriver/webdriver_test.py
```diff
@@ -250,6 +250,75 @@ def exceptionCallback(request, uri, headers):
...
+        except ValueError:
+            assert True
+
+    @httpretty.activate
+    def test_invalid_method(self):
+        driver = ios_w3c_driver()
+        try:
+            driver.add_command(method='error', url='session/$sessionId/path/to/custom/url', name='test_command')
+            assert False, 'Should raise ValueError'
+        except ValueError:
```
mykola-mokhnach, line 319: Pytest has a handy shortcut for this type of tests:

```
with pytest.raises(ValueError):
    driver.add_command(method='error', url='session/$sessionId/path/to/custom/url', name='test_command')
```

appium/webdriver/webdriver.py
```diff
@@ -356,6 +357,94 @@ def switch_to(self) -> MobileSwitchTo:
...
+                    url='session/$sessionId/path/to/custom/url',
+                    name='test_command'
+                )
+
+        """
+        if name in self.command_executor._commands:
+            raise ValueError("{} is already defined".format(name))
+
+        if not isinstance(method, CommandMethod):
+            raise ValueError("'{}' is invalid. Valid method is in '{}'.".format(method, CommandMethod.__name__))
```
mykola-mokhnach, line 393: ...should be one of ...

mykola-mokhnach (approved):
It would be nice to mention this feature in the driver documentation somewhere

jlipps:
this is fantastic @KazuCocoa! one question: so let's say I wanted to create a client sidecar module for people using my custom driver. I could instruct my users to do something like this:

```py
from appium import webdriver
from mydriver_client import update_driver

driver = webdriver.Remote(...)
update_driver(driver)
result = driver.my_custom_command(args)
```

or, if I wanted to wrap the entire lib, I could do:

```py
from mydriver_client import webdriver
driver = webdriver.Remote(...)
result = driver.my_custom_command(args)
```

as the client extension library author I could easily provide both interfaces. The latter interface is more concise, but the former interface could be used if there are multiple client extensions that need to be applied to the driver.

i guess one other thought I have: is there any reason for us to standardize the extension interface to make it easy for extension authors and users? so we could do something like this from an end user perspective:

```py
from appium import webdriver
import client_ext1
import client_ext2

driver = webdriver.Remote(..., extensions=[client_ext1, client_ext2])
driver.custom_command_from_ext1()
driver.custom_command_from_ext2()
```

so that our appium client code knows how to extract command information from the 3rd party client extensions.

what do you think @KazuCocoa @mykola-mokhnach?

KazuCocoa:
ah, seems good idea > `extensions=[client_ext1, client_ext2]`
Let me try it out

jlipps:
Yeah this way all client extensions can have a simple template to follow. For example we could publish the find_by_image commands a built-in extension module that people could just copy

```
from appium import webdriver
from appium.extensions import images

driver = webdriver.Remote(..., extensions=[images])
```

this mirrors the fact that the image functionality is a plugin now.

PR #615, opened 2021-06-23
feat: Add command with `setattr`

KazuCocoa:
As https://github.com/appium/python-client/pull/608#issuecomment-856307740 and after I considered better way, I think we can add a new method dynamically with `setattr`. Then, the new command can call as `driver.new_command`. It should be better than call a method as string as the previous one.

I haven't updated docstring.


appium/webdriver/webdriver.py
```diff
@@ -114,6 +114,43 @@ def _make_w3c_caps(caps: Dict) -> Dict[str, List[Dict[str, Any]]]:
...
 
+class ExtensionBase:
+    """
+    Used to define an extension command.
+
+    TODO:
+    add examples
+    """
+
+    def __init__(self, execute: Callable):
```
mykola-mokhnach, line 125: Python also allows to make typing more precise with Generics, for example Callable[str, str] defines a method, which returns str and accepts one param of type str

appium/webdriver/webdriver.py
```diff
@@ -114,6 +114,43 @@ def _make_w3c_caps(caps: Dict) -> Dict[str, List[Dict[str, Any]]]:
...
+        self._execute = execute
+
+    def execute(self, parameters: Any) -> Any:
```
mykola-mokhnach, line 128: are parameters required?

appium/webdriver/webdriver.py
```diff
@@ -114,6 +114,43 @@ def _make_w3c_caps(caps: Dict) -> Dict[str, List[Dict[str, Any]]]:
...
+        return self._execute(self.method_name(), parameters)
+
+    def method_name(self) -> str:
+        """
+        Expected to return a method name.
+        This name will be able to call as a driver method.
```
mykola-mokhnach, line 134: will be available for calling

appium/webdriver/webdriver.py
```diff
@@ -114,6 +114,43 @@ def _make_w3c_caps(caps: Dict) -> Dict[str, List[Dict[str, Any]]]:
...
+
+        Returns:
+            'str' The method name.
+        """
+        raise NotImplementedError()
+
+    def command_wrapper(self) -> Any:
```
mykola-mokhnach, line 141: some examples would be nice to have for the next two methods

appium/webdriver/webdriver.py
```diff
@@ -176,6 +214,16 @@ def __init__(
         By.IMAGE = MobileBy.IMAGE
         By.CUSTOM = MobileBy.CUSTOM
 
+        for extension in extensions:
+            instance = extension(self.execute)
+            if hasattr(WebDriver, instance.method_name()):
+                logger.warn(
+                    '{} is alreadt defined. The new one is overriding the old one.'.format(instance.method_name())
```
mykola-mokhnach, line 221: already defined

mykola-mokhnach: consider using f-strings instead

appium/webdriver/webdriver.py
```diff
@@ -176,6 +214,16 @@ def __init__(
...
+                )
+            setattr(WebDriver, instance.method_name(), instance.command_wrapper)
```
mykola-mokhnach, line 223: should we actually invoke `command_wrapper` here? I'm not sure about what this method is supposed to return

appium/webdriver/webdriver.py
mykola-mokhnach: overriding existing command might cause unexpected results. I would rather throw an exception here

test/unit/webdriver/webdriver_test.py
```diff
@@ -254,70 +260,73 @@ def exceptionCallback(request, uri, headers):
 
     @httpretty.activate
     def test_add_command(self):
-        driver = ios_w3c_driver()
+        class CustomURLCommand(ExtensionBase):
+            def method_name(self):
+                return 'test_command'
+
+            def command_wrapper(self):
+                return self.execute({})['value']
```
mykola-mokhnach, line 268: I'm not quite sure how are we supposed to pass custom arguments here?

appium/webdriver/webdriver.py
KazuCocoa: yea, I'm looking for instance wide inclusion like Ruby's `include`  to raise an error as previous add_command stuff.
`hasattr` seems class or loaded module wide

jlipps (commented):
👀

Pipfile
```diff
@@ -18,6 +18,7 @@ tox = "~=3.23"
 
 httpretty = "~=1.0"
 python-dateutil = "~=2.8"
+types-python-dateutil = "~=0.1"
```
KazuCocoa, line 21: Azure failure pointed to this necessity.

appium/webdriver/webdriver.py
```diff
@@ -114,6 +114,113 @@ def _make_w3c_caps(caps: Dict) -> Dict[str, List[Dict[str, Any]]]:
...
+class ExtensionBase:
+    """
+    Used to define an extension command as driver's methods.
+
+    Example:
+        When you want to add `example_command` which calls a get request to
+        `session/$sessionId/path/to/your/custom/url`.
+
+        1. Defines an extension as a subclass of `ExtensionBase`
+            class CustomYourCommand(ExtensionBase):
```
jlipps, line 126: `YourCustomCommand`

appium/webdriver/webdriver.py
```diff
@@ -176,6 +284,24 @@ def __init__(
         By.IMAGE = MobileBy.IMAGE
         By.CUSTOM = MobileBy.CUSTOM
 
+        self._extensions = extensions
+        for extension in self._extensions:
+            instance = extension(self.execute)
```
jlipps, line 289: we use `instance.method_name()` a lot. should we just make it a variable for readability?

```
name = instance.method_name()
```
or
```
method_name = instance.method_name()
```

jlipps (approved):
awesome!

appium/webdriver/webdriver.py
```diff
@@ -114,6 +114,113 @@ def _make_w3c_caps(caps: Dict) -> Dict[str, List[Dict[str, Any]]]:
...
+
+            # Then, the driver calls a get request to `session/$sessionId/path/to/your/custom/$id/url`
+            # with replacing the `$id` with the given `element.id`
+            driver.custom_method_name(element.id)
+    """
+
+    def __init__(self, execute: Callable[[str, Dict], Dict[str, Any]]):
+        self._execute = execute
+
+    def execute(self, parameters: Any = {}) -> Any:
```
mykola-mokhnach, line 204: Please do not set mutable objects as defaults: https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

mykola-mokhnach: shall we use `*args, **kwargs` notation instead?

KazuCocoa: Changed here as Dict explicitly to match with original execute: https://github.com/SeleniumHQ/selenium/blob/e46dba531896f67bdb93ba8d9e1d64f8b6b75b6f/py/selenium/webdriver/remote/webdriver.py#L380

PR #620, opened 2021-07-20
feat: add satellites in set_location

KazuCocoa:


appium/webdriver/extensions/location.py
```diff
@@ -51,6 +52,7 @@ def set_location(
             longitude: String or numeric value between -180.0 and 180.0
             altitude: String or numeric value (Android real device only)
             speed: String or numeric value larger than 0.0 (Android real devices only)
+            satellites: String or numeric value since Appium 1.22.0
```
mykola-mokhnach, line 55: We could have a more precise documentation string here:
Count of active GPS satellites in range 1..12. Only works for Emulators.

test/unit/webdriver/device/location_test.py
```diff
@@ -33,25 +33,27 @@ def test_toggle_location_services(self):
     def test_set_location_float(self):
         driver = android_w3c_driver()
         httpretty.register_uri(httpretty.POST, appium_command('/session/1234567890/location'))
-        assert isinstance(driver.set_location(11.1, 22.2, 33.3, 23.2), WebDriver)
+        assert isinstance(driver.set_location(11.1, 22.2, 33.3, 23.2, 0), WebDriver)
```
mykola-mokhnach, line 36: zero is not an acceptable value

mykola-mokhnach (approved):

PR #631, opened 2021-08-29
[wip] add test to check log path

KazuCocoa:
to check current behavior.
It seems they changed https://github.com/SeleniumHQ/selenium/blob/6e1307270a861e2433b6c5d9547a915108f79c6b/py/selenium/webdriver/remote/remote_connection.py#L312-L315


KazuCocoa:
3.x was no-se. ✅

Issue #645, opened 2021-10-15
remove find_element_by_XXXX


KazuCocoa:
Other bindings which already had `find_element_by_XXX` were removed. (e.g. Java client. Ruby client did not have like this syntax)
Python client still has them in selenium repo, so no reason to drop only in this client.
But potentially we'll need to drop them.

Let me leave an issue as a note.

Before removing them, we must print deprecated log in each method

KazuCocoa:
https://appium.github.io/python-client-sphinx/webdriver.common.html#webdriver.common.appiumby.AppiumBy
https://appium.github.io/python-client-sphinx/webdriver.common.html#module-webdriver.common.mobileby

PR #657, opened 2021-11-24
chore: add deprecated mark for find_element_by*

KazuCocoa:
Part of https://github.com/appium/python-client/issues/645

It looks like we've recommended to use `find_element_by_*` instead of vanilla `fine_element(xxx)`, so let's print deprecated message first (and for a while)

### DIFF
--- a/appium/webdriver/extensions/search_context/android.py
+++ b/appium/webdriver/extensions/search_context/android.py
@@ -17,6 +17,7 @@
 import json
 from typing import TYPE_CHECKING, Any, List, Optional, TypeVar, Union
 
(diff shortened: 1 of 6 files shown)


mykola-mokhnach (approved):
It would also make sense to deprecate `MobileBy` and rename it to `AppiumBy`, like we did in java client. the reason for it that Appium does not just support mobile platforms and suhc locations don't make much sense for Desktop automation

KazuCocoa:
I see. It makes sense > AppiumBy

Let me add it and update deprecation messages to use it instead of `MobileBy`

PR #659, opened 2021-11-25
feat: add AppiumBy instead of MobileBy

KazuCocoa:
part of https://github.com/appium/python-client/issues/645

### DIFF
--- a/appium/webdriver/common/appiumby.py
+++ b/appium/webdriver/common/appiumby.py
@@ -12,8 +12,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from appium.webdriver.common.appiumby import AppiumBy
+from selenium.webdriver.common.by import By
 
 
-class MobileBy(AppiumBy):
-    pass
+class AppiumBy(By):
+    IOS_PREDICATE = '-ios predicate string'
+    IOS_UIAUTOMATION = '-ios uiautomation'
+    IOS_CLASS_CHAIN = '-ios class chain'
+    ANDROID_UIAUTOMATOR = '-android uiautomator'
+    ANDROID_VIEWTAG = '-android viewtag'
+    ANDROID_DATA_MATCHER = '-android datamatcher'
+    ANDROID_VIEW_MATCHER = '-android viewmatcher'
+    # Deprecated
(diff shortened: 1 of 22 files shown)


appium/webdriver/common/appiumby.py
```diff
@@ -0,0 +1,30 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from selenium.webdriver.common.by import By
+
+
+class AppiumBy(By):
```
mykola-mokhnach, line 18: I'd prefer it to not be a breaking change. E.g. keep the MobileBy strategy, but simply mark it as deprecated
AppiumBy may then be inherited from MobileBy to avoid unnecessary duplication.

KazuCocoa: `MobileBy.ID` in appium/webdriver/common/mobileby.py (inherits AppiumBy) etc are `str`. `AppiumBy.ID` also `str`. So basically both ID are set as `id` to find_element, for example.
I think they do not have breaking change... something wrong?

Yea, I was finding the way to print deprecation message when `MobileBy.ID` was called or when `MobileBy.ID` was given in find_element. But it was simply `str`.
So currently I've put the deprecation message in the module description.

mykola-mokhnach: Maybe we could use https://www.py4u.net/discuss/182959 (https://pypi.org/project/deprecation/#description) to mark things as deprecated

mykola-mokhnach: Also https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated

KazuCocoa: mm, the deprecation package also does not work for str of class variables?

Yea, I think only documentation can mention it.

KazuCocoa:
mm, it looks like current building docs partially depend on the published python client module.
Will check after pushing a new one (deprecated marks)

Issue #667, opened 2021-12-17
appium-python 2.10 Android phone swipe does not work


tewbie:
## The problem
appium-python-client 2.10  使用driver.swipe() 在安卓手机中，无法拖动元素

## Environment
- Appium version (or git revision) that exhibits the issue: 
- Last Appium version that did not exhibit the issue (if applicable):
- Desktop OS/version used to run Appium:
- Node.js version (unless using Appium.app|exe):
- Mobile platform/version under test: Android 10
- Real device or emulator/simulator: Real device
- Appium CLI or Appium.app|exe:

## Details

I have an operation that uses the driver.swipe() method to drag elements, This method doesn't seem to work in appium-python-client 2.10. After switching to version 1.20, the operation can be performed normally
Turn on the Android pointer position. When using version 2.10, the phone screen shows a sliding track, but elements are not dragged.

## Link to Appium Logs

Create a [GIST](https://gist.github.com) which is a paste of your _full_ Appium logs, and link them here.

## Code To reproduce issue

self.driver.swipe(start_x=start_x, start_y=start_y, end_x=end_x, end_y=end_y, duration=duration)

actions = ActionChains(self.driver)
actions.w3c_actions.pointer_action.move_to_location(start_x, start_y)
actions.w3c_actions.pointer_action.click_and_hold()
actions.w3c_actions.pointer_action.pause(3)
actions.w3c_actions.pointer_action.move_to_location(end_x, end_y)
actions.perform()

KazuCocoa:
So, did not w3c actions work in your case even the pointer location shown correct line?

tewbie:
> 



> 那么，w3c操作是否适用于您的情况，即使指针位置显示正确的行？

Yes,In my operation the swipe method works fine in version 1.20, but using version 2.10 does not work for that operation.   
Same code, but switched appium-Python version
![aaa](https://user-images.githubusercontent.com/51439645/146714966-4a25abdd-07f6-40c3-8ecb-59d7cb00aeaf.png)

KazuCocoa:
As https://github.com/appium/python-client#quick-migration-guide-from-v1-to-v2 , it was because probably the backend had changed to W3C actions.

Does like the below click action work against the icon?
I wonder android API used in w3c did not work to click the icon, or the set of swipe action's w3c actions had an issue.
```
            actions.w3c_actions.pointer_action.move_to(element)
            actions.w3c_actions.pointer_action.pause(duration_sec)
[...]

PR #670, opened 2021-12-24
feat: use 'touch' pointer action

KazuCocoa:
For https://github.com/appium/python-client/issues/667#issuecomment-1000146316

By default, python clients set `mouse` mode. I guess it could be `touch`.


mykola-mokhnach:
I'm not quite sure these shortcuts have to be there. It is probably impossible to make them working for all supported platforms anyway. I'd rather deprecate them and let users build these chains/supply appropriate documentation.

KazuCocoa:
> I'd rather deprecate them and let users build these chains/supply appropriate documentation.

yea, this change told me the old TouchAction should be `touch` input as W3C. So, am thinking to prepare documentation to convert TouchActions to W3C in https://appium.io/docs/en/commands/interactions/actions/ or this repository. (for python, and address these differences as well).

This diff may help as an example in short term, and make this shortcut deprecate as a long term.
(Probably when dropping TouchActions is a good time)

PR #674, opened 2022-01-01
test: update tests to use find_element(by...)

KazuCocoa:
We currently made `find_element_by_xxx` methods as deprecated. It would be nice to all our tests follow it.


mykola-mokhnach (approved):

test/unit/webdriver/search_context/windows_test.py
```diff
@@ -13,7 +13,9 @@
 # limitations under the License.
 
 import httpretty
+from selenium.webdriver.common import by
```
mykola-mokhnach, line 16: I assume this particular was never executed and is invalid. The `-windows uiautomation` locator type is not supported on the server side and was removed from other clients.

KazuCocoa: Yes. this lib also had a deprecation mark.

Issue #688, opened 2022-02-26
selenium 4.1 update


KazuCocoa:
Update https://github.com/appium/python-client/pull/687

Issue #698, opened 2022-03-17
Issues using expected conditions over mobile locators (Selenium 4)


Juanap12:
## The problem

I'm trying to wait for an element to be clickable. This exact same wait does work for our web automation tests, but is showing the following error for mobile elements:

`AttributeError: type object 'Command' has no attribute 'IS_ELEMENT_DISPLAYED'`

This is the wait we're trying to perform

`element = WebDriverWait(parent, timeout, poll_frequency=poll,
                                 ignored_exceptions=[ElementNotVisibleException,
                                                     ElementNotSelectableException,
                                                     StaleElementReferenceException])\
                .until(expected_conditions.element_to_be_clickable(selector))`

"parent" is the driver
"timeout" is a value given in seconds
"selector" is a tuple that has the location strategy and the locator itself. e.g. `('-android uiautomator', 'new UiSelector().textContains("get started")')`

## Environment
- Appium version (or git revision) that exhibits the issue: Currently testing over `selenium>=4.1.3` and `Appium-Python-Client>=2.1.4`
- Desktop OS/version used to run Appium: macOS Monterey
- Node.js version (unless using Appium.app|exe):
- Mobile platform/version under test: Android 10
- Real device or emulator/simulator: emulator
- Appium CLI or Appium.app|exe: 1.22.0

## Details

Worth to mention that if I perform a `driver.find_element(*selector)` it does find the element. It shows the following result `<appium.webdriver.webelement.WebElement (session="a7f5a8a1-4e63-4847-bcc7-265666d7760b", element="00000000-0000-010c-ffff-ffff0000001e")>`. But, if I try to use the expected condition ".element_to_be_clickable" it shows the error mentioned at the beginning

Juanap12:
I've tested using the exact same setup, just changing to selenium 4.0.0 and selenium 4.1.0 library and everything worked as expected. The issue is with selenium >4.1.1

KazuCocoa:
Please use latest appium client that restricts the selenium version as 4.0 https://github.com/appium/python-client/pull/686
https://github.com/appium/python-client/issues/688 is an issue to update this client for selenium 4.1+

KazuCocoa:
ah... your case was using latest appium client with selenium 4.1.1's WebDriverWait.  (not installed with appium cleint)

Juanap12:
[...]

PR #701, opened 2022-03-25
feat: add non-w3c but still need commands

KazuCocoa:
Fixes https://github.com/appium/python-client/issues/688 https://github.com/appium/python-client/issues/698

https://github.com/appium/ruby_lib_core/blob/master/lib/appium_lib_core/common/command.rb should be a good reference.

### DIFF
--- a/appium/webdriver/extensions/ime.py
+++ b/appium/webdriver/extensions/ime.py
@@ -34,7 +34,7 @@ class IME(CanExecuteCommands):
         Returns:
             :obj:`list` of :obj:`str`: The available input methods for an Android device
         """
-        return self.execute(Command.GET_AVAILABLE_IME_ENGINES, {})['value']
+        return self.execute(Command.GET_AVAILABLE_IME_ENGINES, {}).get('value', [])
 
     def is_ime_active(self) -> bool:
         """Checks whether the device has IME service active.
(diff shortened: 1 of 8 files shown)


KazuCocoa:
`WebDriverWait` is also in some test code. They passed.

KazuCocoa:
let me merge and publish it as a new version

appium/webdriver/extensions/ime.py
```diff
@@ -34,7 +34,7 @@ def available_ime_engines(self) -> List[str]:
         Returns:
             :obj:`list` of :obj:`str`: The available input methods for an Android device
         """
-        return self.execute(Command.GET_AVAILABLE_IME_ENGINES, {})['value']
+        return self.execute(Command.GET_AVAILABLE_IME_ENGINES, {}).get('value', [])
```
mykola-mokhnach, line 37: I would rather keep it as it was. A user might get an incorrect response if the value is missing instead of an error, which is worse

mykola-mokhnach: same below

KazuCocoa: okay. this was just to fix lint in VSCode

PR #706, opened 2022-03-29
revert: pylint

KazuCocoa:


appium/webdriver/extensions/location.py
```diff
@@ -80,7 +80,7 @@ def location(self) -> Dict[str, float]:
                 - longitude (float)
                 - altitude (float)
         """
-        return self.execute(Command.GET_LOCATION).get('value', {})
+        return self.execute(Command.GET_LOCATION)['value']  # pylint: disable=unsubscriptable-object
```
mykola-mokhnach, line 83: this might be an issue of pylint itself, because self.execute is known to return Dict
🤷

mykola-mokhnach (approved):

mykola-mokhnach:
I'd say for now it would be ok to also unlock selenium dependency, so minor versions are included as well (e.g. "<5.0"). It is usually too painful to bump them manually every time. We could always limit the max version again if some major issue pops up

PR #707, opened 2022-04-01
support send keys to active_element

Chris8711:
add a mixin class ActiveElement
support function get_active_element  and send_keys_to_active_element

### DIFF
--- a/appium/webdriver/extensions/active_element.py
+++ b/appium/webdriver/extensions/active_element.py
@@ -32,12 +32,11 @@ class ActiveElement(CanExecuteCommands):
 
             `WebElement`: web element
         """
-        try:
-            return self.execute(W3C_Command.W3C_GET_ACTIVE_ELEMENT)['value']
-        except NoSuchElementException:
-            return None
+        return self.execute(W3C_Command.W3C_GET_ACTIVE_ELEMENT)['value']
 
     def send_keys_to_active_element(self, value: str):
-        element = self.get_active_element()
-        if element:
+        try:
+            element = self.get_active_element()
             return element.clear().send_keys(value)
+        except NoSuchElementException:
+            return None
--- a/appium/webdriver/extensions/active_element.py
+++ b/appium/webdriver/extensions/active_element.py
@@ -1,0 +1,42 @@
+#!/usr/bin/env python
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


appium/webdriver/extensions/active_element.py
```diff
@@ -0,0 +1,43 @@
...
+        It usually used after tapping a point and active the element.
+        And use the element.
+
+        Returns:
+
+            `WebElement`: web element
+        """
+        try:
+            return self.execute(W3C_Command.W3C_GET_ACTIVE_ELEMENT)['value']
+        except NoSuchElementException:
```
mykola-mokhnach, line 37: it is ok to throw the exception from this method: https://www.w3.org/TR/webdriver/#dfn-get-active-element

appium/webdriver/extensions/active_element.py
```diff
@@ -0,0 +1,43 @@
...
+            return None
+
+    def send_keys_to_active_element(self, value: str):
```
mykola-mokhnach, line 40: I don't see any actual point in this helper method. also, it is not a part of the spec

Chris8711: the driver.tap([positions]) will actually  click some element

Chris8711: I use other tool the get the position on the screenshot.
and use appium client to tap or click the screen.
if some element was tapped. Then I can use get_active_element to do other things

mykola-mokhnach:
I can observe that selenium python lib already has a wrapper for active element:

```
class SwitchTo:
    def __init__(self, driver):
        import weakref
        self._driver = weakref.proxy(driver)

    @property
    def active_element(self) -> WebElement:
        """
        Returns the element with focus, or BODY if nothing has focus.

        :Usage:
            ::

                element = driver.switch_to.active_element
        """
        return self._driver.execute(Command.W3C_GET_ACTIVE_ELEMENT)['value']
```

Wouldn't this be a duplicate then?

Chris8711:
> 



> I can observe that selenium python lib already has a wrapper for active element:
> 
> ```
> class SwitchTo:
>     def __init__(self, driver):
>         import weakref
>         self._driver = weakref.proxy(driver)
> 
>     @property
>     def active_element(self) -> WebElement:
>         """
>         Returns the element with focus, or BODY if nothing has focus.
> 
>         :Usage:
>             ::
> 
>                 element = driver.switch_to.active_element
>         """
>         return self._driver.execute(Command.W3C_GET_ACTIVE_ELEMENT)['value']
> ```
> 
> Wouldn't this be a duplicate then?

yes. I haven't noticed that.
Thank you

Chris8711:
this is duplicated

KazuCocoa:
Btw, sending keys to an active element is now W3C's send_keys action.
https://github.com/appium/appium/issues/16419#issuecomment-1031764419

Get an active element -> calling send_keys to the element is also the same, but the w3c action's one is old send_keys_active_element (`[:post, 'session/:session_id/keys']`) endpoint.

PR #724, opened 2022-06-13
chore: Update comments to locator patches

VladimirPodolyan:
Hi! I faced with an issue while using `appium.find_element` with locators, that I used for my selenium tests. Is it possible to uncomment [that block](https://github.com/appium/python-client/pull/724/files#diff-646299ef4dd5d3b0e5d58632c99871ba450a1481e017654f4e95b11a2847c945R340) and use him in `web view` context only ?

### DIFF
--- a/appium/webdriver/webdriver.py
+++ b/appium/webdriver/webdriver.py
@@ -337,18 +337,18 @@ class WebDriver(
             `appium.webdriver.webelement.WebElement`: The found element
 
         """
-        # TODO: If we need, we should enable below converter for Web context
-        # if by == By.ID:
-        #     by = By.CSS_SELECTOR
-        #     value = '[id="%s"]' % value
-        # elif by == By.TAG_NAME:
-        #     by = By.CSS_SELECTOR
-        # elif by == By.CLASS_NAME:
-        #     by = By.CSS_SELECTOR
-        #     value = ".%s" % value
-        # elif by == By.NAME:
-        #     by = By.CSS_SELECTOR
-        #     value = '[name="%s"]' % value
+        if self.context != 'NATIVE_APP':
+            if by == By.ID:
+                by = By.CSS_SELECTOR
+                value = '[id="%s"]' % value
+            elif by == By.TAG_NAME:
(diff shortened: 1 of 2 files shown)


mykola-mokhnach (changes requested):
see the above comment

VladimirPodolyan:
@mykola-mokhnach Hi, and thank you for answer. 
So, maybe it's possible to pass that as class argument **for both classes**? 

For WebDriver:
```python
class WebDriver(...):
    def __init__(command_executor, desired_capabilities, ..., selenium_compatibility)
        ...
        self.selenium_compatibility = selenium_compatibility

    def find_element():
            if self.selenium_compatibility:
                if by == By.ID:
                    ...
```
For WebElement (need to define `__init__`)
```python
class WebElement(AppiumWebElementSearchContext):
    def __init__(self, selenium_compatibility=False):
        self.selenium_compatibility = selenium_compatibility

    def find_element():
            if self.selenium_compatibility:
                if by == By.ID:
                    ...
```

Or only for `WebDriver`, even you don't accept the specifying the `__init__` in WebElement

VladimirPodolyan:
@mykola-mokhnach Understood. But what is the final decision here? You don't want to have this "Selenium workaround" in your own code, but it's commented right now. Can I just delete this "comment section" in this PR and do a self workaround in my side?

appium/webdriver/webdriver.py
```diff
@@ -337,7 +337,8 @@ def find_element(self, by: str = AppiumBy.ID, value: Union[str, Dict] = None) ->
             `appium.webdriver.webelement.WebElement`: The found element
 
         """
-        # TODO: If we need, we should enable below converter for Web context
+        # Let's keep it for future support of the selenium workaround
```
mykola-mokhnach, line 340: I'd rather change the comment to `We prefer to patch locators in the client code. Checking current context every time a locator is accessed could significantly slow down tests`

mykola-mokhnach (approved):

KazuCocoa (approved):

PR #744, opened 2022-07-01
fix: Backwards compatible behaviour of swipe and scroll in action_helpers

### LINKED ISSUE #743: Suspicious functionality change in driver.swipe() and driver.scroll() duration argument
## The problem

Before the ActionChains were used (Appium 1.20.x, Python Client 1.x.x) as they are now written in action_helpers.py we had used the following method to swipe the screen open from screen lock state:

`driver.swipe(x1, y1, x2, y2, duration=3000)`

This gave a slow swipe that took 3 seconds to move from x1,y1 to x2,y2.

When updating to Appium 2.0 and the latest Appium Python Client, the swipe did not work properly anymore.

If I understand the code properly in action_helpers.py, the functionality now uses the duration argument as pause between the move from x1,y1 to x2,y2:

`actions.w3c_actions.pointer_action.pause(duration / 1000)`

But the actual swipe action is very fast, in my case with a three second pause before the actual fast swipe action!

To get the swipe to work as before, I made a copy of the swipe method from action_helpers.py and changed like this:

    def swipe(self: T, start_x: int, start_y: int, end_x: int, end_y: int, duration: int = 0) -> T:
        actions = ActionChains(self)
        actions.w3c_actions = ActionBuilder(self, mouse=PointerInput(interaction.POINTER_TOUCH, "touch"), duration=duration)
        actions.w3c_actions.pointer_action.move_to_location(start_x, start_y)
        actions.w3c_actions.pointer_action.pointer_down()
        actions.w3c_actions.pointer_action.move_to_location(end_x, end_y)
        actions.w3c_actions.pointer_action.release()
        actions.perform()

**Notice that I have moved the duration parameter to the ActionBuilder constructor and removed the pause completely.**

So I am proposing to implement the swipe() and scroll() like this and if someone requires an actual pause between the move, then I suggest adding an additional pause argument to the method signature.

Also I would like to see long_press(element: WebElement, timeout: int) to be added to action_helpers.py, where the pause would actually use the timeout value.

Thanks! 

## Environment
Appium version: 2.0.0Beta40
Appium Python Client version 2.6.0
OS version: Linux Ubuntu 20.04
Mobile platform: Android (real device)

jatalahd:
- Fixed handling the duration argument in swipe() and scroll() helpers
- Functionality is now the same as in older versions using TouchActions

Fixes #743

### DIFF
--- a/appium/webdriver/extensions/action_helpers.py
+++ b/appium/webdriver/extensions/action_helpers.py
@@ -42,18 +42,20 @@ class ActionHelpers:
         Returns:
             Union['WebDriver', 'ActionHelpers']: Self instance
         """
-
         # XCUITest x W3C spec has no duration by default in server side
         if duration is None:
             duration = 600
 
+        touch_input = PointerInput(interaction.POINTER_TOUCH, "touch")
+
         actions = ActionChains(self)
-        # duration given directly as argument, as the default was set above
-        actions.w3c_actions = ActionBuilder(self, mouse=PointerInput(interaction.POINTER_TOUCH, "touch"), duration=duration)
+        actions.w3c_actions = ActionBuilder(self, mouse=touch_input)
 
         # https://github.com/SeleniumHQ/selenium/blob/3c82c868d4f2a7600223a1b3817301d0b04d28e4/py/selenium/webdriver/common/actions/pointer_actions.py#L83
         actions.w3c_actions.pointer_action.move_to(origin_el)
         actions.w3c_actions.pointer_action.pointer_down()
+        # setup duration for the second move only, assuming duration always has atleast default value
+        actions.w3c_actions = ActionBuilder(self, mouse=touch_input, duration=duration)
         actions.w3c_actions.pointer_action.move_to(destination_el)
         actions.w3c_actions.pointer_action.release()
         actions.perform()
@@ -143,15 +145,14 @@ class ActionHelpers:
         Returns:
             Union['WebDriver', 'ActionHelpers']: Self instance
         """
-        # if duration is not given, do not define it as argument to use default from ActionBuilder
-        kw_args = {"mouse": PointerInput(interaction.POINTER_TOUCH, "touch")}
-        if duration > 0:
-            kw_args["duration"] =  duration
+        touch_input = PointerInput(interaction.POINTER_TOUCH, "touch")
 
         actions = ActionChains(self)
-        actions.w3c_actions = ActionBuilder(self, **kw_args)
+        actions.w3c_actions = ActionBuilder(self, mouse=touch_input)
         actions.w3c_actions.pointer_action.move_to_location(start_x, start_y)
         actions.w3c_actions.pointer_action.pointer_down()
+        if duration > 0:
+            actions.w3c_actions = ActionBuilder(self, mouse=touch_input, duration=duration)
         actions.w3c_actions.pointer_action.move_to_location(end_x, end_y)
         actions.w3c_actions.pointer_action.release()
         actions.perform()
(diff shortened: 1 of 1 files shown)


KazuCocoa:
thank you. I will take a look.
Could you fix the lint error in the CI?

mykola-mokhnach:
we probably need to use timedelta type for duration, so then we never have to mix seconds and milliseconds

jatalahd:
I did not find the actual reason for the lint error in the CI logs, but I made a new commit that hopefully goes through. If this also fails the lint, then you need to tell me what to fix.

Also, I made an important improvement to earlier version, where I did not realize that the duration given in the ActionBuilder affects all move-operations. So the first move to the point of pointer_down was also affected with the duration, which was of course wrong. It was not very intuitive to create a second ActionBuilder instance and assign the duration to only that one. But I tested it and the logs were indicating that it is the correct approach.

KazuCocoa (commented):
Left a comment probably lint error, otherwise lgtm.
Actions are stored into `actions` in `touch_input`, so giving a new instance with the same touch_input helps keep actions.

jatalahd:
Fixed the lint error in the latest commit. This is now ready from my point of view.

appium/webdriver/extensions/action_helpers.py
```diff
@@ -42,32 +42,23 @@ def scroll(self: T, origin_el: WebElement, destination_el: WebElement, duration:
...
-            actions.w3c_actions.pointer_action.move_to(origin_el)
-            actions.w3c_actions.pointer_action.pointer_down()
-            actions.w3c_actions.pointer_action.pause(duration_sec)
-            actions.w3c_actions.pointer_action.move_to_location(dest_el_rect['x'], dest_el_rect['y'])
-            actions.w3c_actions.pointer_action.release()
-            actions.perform()
+        actions.w3c_actions.pointer_action.move_to(origin_el)
+        actions.w3c_actions.pointer_action.pointer_down()
+        # setup duration for second move only, assuming duration always has atleast default value
+        actions.w3c_actions = ActionBuilder(self, mouse=touch_input, duration=duration)
```
mykola-mokhnach, line 58: I don't understand the logic here - do we just reassign `actions.w3c_actions` property? what is then the point of having previous calls?

KazuCocoa: Previous actions are in `touch_input`, so this one just updates `duration`.
`actions.w3c_actions.pointer_action._duration = duration` can be the alternative of the `ActionBuilder(self, mouse=touch_input, duration=duration)` to append actions in `touch_input` with the duration.

```
>>> touch_input
<selenium.webdriver.common.actions.pointer_input.PointerInput object at 0x108c750f0>
>>> touch_input.actions
[{'type': 'pointerDown', 'duration': 0, 'button': 0}, {'type': 'pointerDown', 'duration': 0, 'button': 0}, {'type': 'pointerUp', 'duration': 0, 'button': 0}]
```

KazuCocoa (approved):

appium/webdriver/extensions/action_helpers.py
jlipps: i think you can also call the `create_xxx` methods on touch_input itself, like `touch_input.create_pointer_move()`, right?

jlipps:
for example this is how i draw a line using the python client:

```
def draw_line(driver, sx, sy, ex, ey, duration=250):
    line = ActionBuilder(driver)
    p = line.add_pointer_input(POINTER_TOUCH, "finger1")
    p.create_pointer_move(duration=0, x=sx, y=sy)
    p.create_pointer_down(button=MouseButton.LEFT)
    p.create_pointer_move(x=ex, y=ey, duration=duration)
    p.create_pointer_up(button=MouseButton.LEFT)
    line.perform()
```

seems a bit less confusing.

PR #752, opened 2022-07-30
ci: use npm v6 as a workaround

KazuCocoa:
As same as https://github.com/appium/ruby_lib_core/pull/397, until a new beta will be released to keep CI working

https://github.com/appium/python-client/runs/7590168516 is the current CI I just re-ran yesterday.

### DIFF
--- a/ci-jobs/functional/setup_appium.yml
+++ b/ci-jobs/functional/setup_appium.yml
@@ -3,6 +3,8 @@ steps:
   inputs:
     versionSpec: '16.x'
   displayName: Install Node 16.x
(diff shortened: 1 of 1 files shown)


mykola-mokhnach:
I would rather wait until the original issue is fixed and deployed by @boneskull 
I don't really want to downgrade NPM everywhere to upgrade it again after a while

boneskull:
agree w/ @mykola-mokhnach

KazuCocoa:
beta 42 has the fix. (I've confirmed the fix)

PR #776, opened 2022-09-27
ci: run unit tests on multiple env

KazuCocoa:
trying to run tests on win as well


.github/workflows/unit-test.yml
```diff
@@ -8,10 +8,11 @@ on:
     branches: [ master ]
 
 jobs:
-  test-ubuntu:
-    runs-on: ubuntu-latest
+  test:
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
+        os: [macos-latest, ubuntu-latest]  # windows-latest does not work yet
```
KazuCocoa, line 15: Win requires shell script update, so not included yet

mykola-mokhnach: xcuitest driver tests only make sence on mac platform
Also AFAIK our Android emulator setup script did not work on Windows

KazuCocoa: yea, this is just for python unit tests, not functional ones. adding instances may be no impact on these small tests, so tried to add Win

mykola-mokhnach: what is the reason we need to run unit tests on multiple platforms? Do you think this might uncover extra issues? I understand win and *nix do have major differences. But linux and macos are quite similar from the client perspective - same posix environment

KazuCocoa: i thought initially python client would have platform specific behavior such as managing appium server. Then, running it as functional test on multiple platform may take time, so it would be nice to check something in unit test. I imagined something like building CLI depending on platform

KazuCocoa:
(anyway, this did not work on Win)

PR #781, opened 2022-10-01
chore: allow selenium client v4.5

KazuCocoa:
https://github.com/appium/python-client/pull/780 tests seems good except for the test host machine's flakiness


KazuCocoa:
oh, this is not necessary -> https://github.com/appium/python-client/blob/master/setup.py#L51

PR #793, opened 2022-10-29
feat: make the UA format with same as other clients

KazuCocoa:
Current UA in this library is `appium/python 0.49 (selenium/3.141.0 (python linux)))`.
Like Java and Ruby clients, this lib also can follow a similar format with `/version`. e.g. https://github.com/appium/java-client/blob/8ebb5c320a1b98fbc14ac2f2c9cec1447db76c34/src/main/java/io/appium/java_client/AppiumUserAgentFilter.java#L46-L47 So, it will be `appium/0.49 (selenium/3.141.0 (python linux)))`


mykola-mokhnach (approved):

mykola-mokhnach:
@KazuCocoa could you please rebase and rerun tests?

KazuCocoa:
oh, test failures came from the desired_capabilities issue

PR #803, opened 2022-12-05
Flake8 black changes

marksmayo:
Ran the black formatter over the codebase. There are some wins here in that it fixes a lot of style issues, tidies imports and consistently lays out the codebase.

Ideally this could be added to github actions or pre_commit.

All tests were re-run and no additional failures identified.


mykola-mokhnach:
black formatting is already applied to the codebase
I don't see any reason in changing single quote to double quotes, for example

KazuCocoa:
https://github.com/appium/python-client/blob/91a9d6701430599637bb6895f9ffb078ee4bd052/Makefile#L14 is current our back command, so this is not necessary. Thanks anyway :)

PR #809, opened 2022-12-13
chore: remove duplicated clean command

KazuCocoa:
- clear command exists in the selenium client, so it is safe to remove it from this library. Maybe.... the original motivation to add this endpoint was to add the command for MJSONWP.
    - https://github.com/SeleniumHQ/selenium/blob/084c3b16bb11d625536141bc441ec95a8ee05bb7/py/selenium/webdriver/remote/webelement.py#L114-L116
    - https://github.com/SeleniumHQ/selenium/blob/084c3b16bb11d625536141bc441ec95a8ee05bb7/py/selenium/webdriver/remote/remote_connection.py#L204
    - The removed one should not be an issue since at least, v4, is the below at least


mykola-mokhnach (approved):

mykola-mokhnach:
@KazuCocoa I think I know the reason why this override has been present there: https://discuss.appium.io/t/clear-is-not-working-anymore-what-is-the-alternative/39421/5

The original API in selenium returns None while the overridden one returns the self instance, so chaining could be possible

PR #856, opened 2023-05-01
refactor: Move driver-specific commands to use extensions (part1)

mykola-mokhnach:


KazuCocoa:
I haven't reviewed fully yet, but maybe after merging all parts, probably we should bump the major version (at least minor version) as fallback with try/except. Cloud users for example will have additional command delay forcefully, it would increase users' execution time. it might be not patch version change as users.
Then, probably similar to https://github.com/appium/python-client#quick-migration-guide-from-v1-to-v2 note may be needed. I'll create an issue to do it.

mykola-mokhnach:
I've also added several helpers to remember the absence of mobile methods, so the command might be slower just during the very first call. All the further calls should be fast as they don't send unnecessary server requests

appium/webdriver/webdriver.py
```diff
@@ -497,9 +500,33 @@ def orientation(self, value: str) -> None:
...
+
+    def mark_extension_absence(self, ext_name: str) -> 'WebDriver':
+        """
+        Marks the given extension as absent for the given driver instance.
+        This API is designed for private usage.
+
+        :param ext_name: extension name
+        :return: self instance for chaining
+        """
+        self._absent_extensions.add(ext_name)
```
KazuCocoa, line 527: Can we leave a log here? Perhaps... debug or info?

The main usage is for us. When we get reports by users even if they use the latest driver versions, we maybe should add new extension commands for drivers. Such logs will help to find them.

KazuCocoa (approved):
otherwise lgtm

appium/webdriver/webdriver.py
mykola-mokhnach: added a log

KazuCocoa (approved):

chrisgraham24:
Hi,

Is the following error in v2.10.0 related to this change?

E       selenium.common.exceptions.WebDriverException: Message: Unknown mobile command "pressKey". Only shell,execEmuConsoleCommand,dragGesture,flingGesture,doubleClickGesture,clickGesture,longClickGesture,pinchCloseGesture,pinchOpenGesture,swipeGesture,scrollGesture,scrollBackTo,scroll,viewportScreenshot,viewportRect,deepLink,startLogsBroadcast,stopLogsBroadcast,acceptAlert,dismissAlert,batteryInfo,deviceInfo,getDeviceTime,changePermissions,getPermissions,performEditorAction,startScreenStreaming,stopScreenStreaming,getNotifications,listSms,type,sensorSet,pushFile,pullFile,pullFolder,deleteFile,isAppInstalled,queryAppState,activateApp,removeApp,terminateApp,installApp,clearApp,startActivity,startService,stopService,broadcast,getContexts,installMultipleApks,unlock,refreshGpsCache,startMediaProjectionRecording,isMediaProjectionRecordingRunning,stopMediaProjectionRecording commands are supported.

Issue #865, opened 2023-05-15
Issue with current_activity in 2.10.0


kakooloukia:
## Breaking change in 2.10.0
**WebDriverException: Message: Unknown mobile command "getCurrentActivity"**

## Environment
- Appium-Python-Client-2.10.0
- Android 10
- UiAutomator2

## Details

![image](https://github.com/appium/python-client/assets/52101924/617b46de-7fa8-4d1a-a556-576cbd9e867b)

## Breaking Change:
![image](https://github.com/appium/python-client/assets/52101924/2241868b-06dd-49b0-b72a-b85c29cc8c6a)
source [appium/webdriver/extensions/android/activities.py](https://github.com/appium/python-client/compare/v2.9.0...master#diff-5f03459b034110b120fd24abae517d2bf9c1e93faa5d55e31dae9117b5309849)

## Code To reproduce issue
open a driver with android UiAutomator on an android device and run `driver.current_activity`

kakooloukia:
The fallbacks is not working : https://github.com/appium/python-client/blob/a369b6fa9efee502a31349a6a84dd1cdd9c5d070/appium/webdriver/webdriver.py#L505-L517

`self._absent_extensions` is empty, so `return self` instead `UnknownMethodException`

kakooloukia:
3 Solutions:
- feed `self._absent_extensions` during the driver creation
- remove `UnknownMethodException` exception filter
- add and use `self._activated_extensions`

chrisgraham24:
Can we expect a fix for this? Or do we need to just use 2.9.0?

KazuCocoa:
Please use 2.9.0 right now

KazuCocoa:
2.10.1 has the fix.

PR #868, opened 2023-05-19
fix UnknownCommandError

KazuCocoa:
for https://github.com/appium/python-client/issues/865

Still initial implementation.

### DIFF
--- a/appium/protocols/webdriver/can_remember_extension_presence.py
+++ b/appium/protocols/webdriver/can_remember_extension_presence.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import TypeVar
+from typing import TypeVar, Union
 
 from ..protocol import Protocol
 
@@ -25,3 +25,8 @@ class CanRememberExtensionPresence(Protocol):
 
     def mark_extension_absence(self: T, ext_name: str) -> T:
         ...
+
+    def is_missing_command(self: T, error_msg: Union[str, None]) -> bool:
+        if error_msg is None:
+            return False
+        return 'unknown mobile command' in error_msg.lower()
--- a/appium/webdriver/extensions/android/activities.py
+++ b/appium/webdriver/extensions/android/activities.py
@@ -15,7 +15,7 @@
 import warnings
 from typing import TYPE_CHECKING, cast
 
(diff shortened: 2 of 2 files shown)


appium/webdriver/extensions/android/activities.py
```diff
@@ -79,9 +79,11 @@ def current_activity(self) -> str:
         ext_name = 'mobile: getCurrentActivity'
         try:
             return self.assert_extension_exists(ext_name).execute_script(ext_name)
-        except UnknownMethodException:
-            # TODO: Remove the fallback
-            return self.mark_extension_absence(ext_name).execute(Command.GET_CURRENT_ACTIVITY)['value']
+        except WebDriverException as e:
+            if isinstance(e, UnknownMethodException) or self.is_missing_command(e.msg):
```
mykola-mokhnach, line 83: could it be the server throws anything else other than UnknownMethodException in case of missing method?

appium/protocols/webdriver/can_remember_extension_presence.py
```diff
@@ -25,3 +25,8 @@ def assert_extension_exists(self: T, ext_name: str) -> T:
 
     def mark_extension_absence(self: T, ext_name: str) -> T:
         ...
+
+    def is_missing_command(self: T, error_msg: Union[str, None]) -> bool:
```
mykola-mokhnach, line 29: protocols must not contain method implementations

KazuCocoa: yea, i noticed the implementation was in webdriver.py

appium/webdriver/extensions/android/activities.py
mykola-mokhnach: This must be a bug in the client itself. I believe it does not properly map the exception raised by the server because this works as expected with Java

mykola-mokhnach: No worries, I have already prepared a fix: https://github.com/appium/python-client/pull/869

Thanks for pointing this out, I did not know it's so broken. Seems like the actual error handling code in selenium client has not been updated since ages and only supports responses in JWP format

KazuCocoa: thanks, gotcha. Then this pr is no longer needed

PR #874, opened 2023-05-27
docs: address version management recommendation in the readme

KazuCocoa:
To prevent unexpected changes that could break this client's functionality.
Actually, the last break was by a patch version, but maybe it is reasonable to set the limit in the minor version for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,13 @@ download and unarchive the source tarball (Appium-Python-Client-X.X.tar.gz).
     python setup.py install
     ```
 
+> **Note**
+> The Appium Python Client depends on [Selenium Python binding](https://pypi.org/project/selenium/), thus
+> the Selenium Python binding update might affect the Appium Python Client.
+> For exampple, some changes in the Selenium binding could break the Appium client.
+> To reduce such unexpected behavior, we strongly recommend you to manage dependencies
+> with version management tool such as Pipenv and Pipfile.
+
 ## Usage
 
 The Appium Python Client is fully compliant with the WebDriver Protocol
--- a/README.md
+++ b/README.md
@@ -84,12 +84,22 @@ download and unarchive the source tarball (Appium-Python-Client-X.X.tar.gz).
     python setup.py install
     ```
 
+## Compatibility Matrix
+
+|Appium Python Client| Selenium binding|
+|----|----|
+|`2.10.0`+ |`4.1.0`+ |
+|`2.2.0` - `2.9.0` |`4.1.0` - `4.9.0` |
+|`2.0.0` - `2.1.4` |`4.0.0` |
+|`1.1.0` and below|`3.x`|
+
+The Appium Python Client depends on [Selenium Python binding](https://pypi.org/project/selenium/), thus
+the Selenium Python binding update might affect the Appium Python Client behavior.
+For exampple, some changes in the Selenium binding could break the Appium client.
+
 > **Note**
-> The Appium Python Client depends on [Selenium Python binding](https://pypi.org/project/selenium/), thus
-> the Selenium Python binding update might affect the Appium Python Client.
-> For exampple, some changes in the Selenium binding could break the Appium client.
-> To reduce such unexpected behavior, we strongly recommend you to manage dependencies
-> with version management tools such as Pipenv and requirements.txt.
+> We strongly recommend you to manage dependencies with version management tools such as Pipenv and requirements.txt
+> to keep compatible version combinations.
 
 ## Usage
 
--- a/README.md
+++ b/README.md
@@ -103,41 +103,102 @@ As a base for the following code examples, the following sets up the [UnitTest](
 environment:
 
 ```python
-# Android environment
+# Python/Pytest
+import pytest
+
 from appium import webdriver
 # Options are only available since client version 2.3.0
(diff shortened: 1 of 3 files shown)


mykola-mokhnach:
is it chore or ci?

KazuCocoa:
Removed the ci change from this pr

mykola-mokhnach:
in java we just set the minimum version but don't limit the maximum one 

in general, it has been already confirmed that selenium does not follow semver, so they may introduce breaking changes even in minor releases. For us this means we must either strictly limit the lib version and then manually release a new version of python client when a new selenium is releases and tested for the compatibility. Or just acknowledge that we are limited in resources and we cannot really cover the testing for each version change..

Based on the above statement my proposal would be to follow the same strategy we have in java:
- the selenium version is set to a range where the minimum one is the recent tested/fixed one and the maximum one is the next major
- we allow selenium client updates without further validations. If a compatibility issue happens we fix it, bump the miniumum supported version (if necessary) and release the new version of the python client
- users can always lock/override selenium's version in their projects as a possible workaround

setup.py
```diff
@@ -48,5 +50,5 @@
         'Topic :: Software Development :: Quality Assurance',
         'Topic :: Software Development :: Testing',
     ],
-    install_requires=['selenium ~= 4.1'],
+    install_requires=['selenium >= 4.1, < 4.10'],
```
mykola-mokhnach, line 53: not sure why this version should be set here. I though Pipfile is the only source of truth

KazuCocoa: Hm, Pipfile is only in this repository, I thought it was for dev purpose. The package does not include it

mykola-mokhnach: Yes, you are right. Sorry for the confusion: https://gist.github.com/dpboard/887b74f242605c3a409b90f0cf706531

Although I am still not quite convinced about version locking strategy

KazuCocoa:
Correct. Hm, actually maybe what we could as our best is to give users to keep local files in their eng. Let me modify this PR to add acknowledgment in the readme

setup.py
KazuCocoa: (sorry, my another comment was reflected slowly)

KazuCocoa:
UPdated

setup.py
```diff
@@ -38,6 +38,8 @@
         'Programming Language :: Python :: 3.7',
         'Programming Language :: Python :: 3.8',
         'Programming Language :: Python :: 3.9',
+        'Programming Language :: Python :: 3.10',
+        'Programming Language :: Python :: 3.11',
```
mykola-mokhnach, line 42: shall we change the `install_requires` limit to `< 5.0`?

README.md
```diff
@@ -84,6 +84,13 @@ download and unarchive the source tarball (Appium-Python-Client-X.X.tar.gz).
     python setup.py install
     ```
 
+> **Note**
+> The Appium Python Client depends on [Selenium Python binding](https://pypi.org/project/selenium/), thus
+> the Selenium Python binding update might affect the Appium Python Client.
+> For exampple, some changes in the Selenium binding could break the Appium client.
+> To reduce such unexpected behavior, we strongly recommend you to manage dependencies
+> with version management tool such as Pipenv and Pipfile.
```
mykola-mokhnach, line 92: Pipfile is just  a config file. The tool itself is called Pipenv.

README.md
```diff
@@ -84,6 +84,13 @@ download and unarchive the source tarball (Appium-Python-Client-X.X.tar.gz).
     python setup.py install
     ```
 
+> **Note**
+> The Appium Python Client depends on [Selenium Python binding](https://pypi.org/project/selenium/), thus
+> the Selenium Python binding update might affect the Appium Python Client.
+> For exampple, some changes in the Selenium binding could break the Appium client.
+> To reduce such unexpected behavior, we strongly recommend you to manage dependencies
```
mykola-mokhnach, line 91: We probably need to explicitly point out the source for a compatible selenium version. Perhaps, maintaining a table like https://github.com/appium/java-client#compatibility-matrix would also make sense

setup.py
KazuCocoa: Existing `selenium ~= 4.1` limits the version up to 4.x, so it is not necessary

README.md
KazuCocoa: Sounds good. need to recall past versions, so let me prepare as another pr

KazuCocoa: -> https://github.com/appium/python-client/pull/876
The version mapping is simpler than Java client's dependencies, so it currently only has a matrix table.

KazuCocoa: Moved into this pr

README.md
```diff
@@ -84,6 +84,23 @@ download and unarchive the source tarball (Appium-Python-Client-X.X.tar.gz).
     python setup.py install
     ```
 
+## Compatibility Matrix
+
+|Appium Python Client| Selenium binding|
+|----|----|
+|`2.10.0`+ |`4.1.0`+ |
```
mykola-mokhnach, line 91: should it be 4.9.1+ ?

KazuCocoa: it worked with older versions, so currently this section does not limit the version. If we need to limit the selenium version, the setup.py will also need to limit it

Issue #878, opened 2023-06-08
Appium-Python-Client 2.10.1 is broken with selenium 4.10.0


ikseek:
## The problem
Any attempt to instantiate remote WebDriver triggers a `TypeError: WebDriver.__init__() got an unexpected keyword argument 'desired_capabilities'`
## Environment
- Appium version: 2.10.1
- Selenium version: 4.10.0
- Desktop OS/version used to run Appium: Mac OS Ventura 13.4

## Details

Looks like the latest Selenium introduced a breaking change in the WebDriver constructor by dropping desired_capabilities arg.

## Code To reproduce issue

```
from appium.webdriver import Remote
Remote()
...
TypeError: WebDriver.__init__() got an unexpected keyword argument 'desired_capabilities'

```

KazuCocoa:
Please try out https://pypi.org/project/Appium-Python-Client/2.10.2/

tking16:
> Please try out https://pypi.org/project/Appium-Python-Client/2.10.2/

@KazuCocoa I've added `Appium-Python-Client==2.10.2` and for some reason I'm getting 
```
[...]

PR #879, opened 2023-06-08
fix: Update the constructor for compatibility with python client 4.10

mykola-mokhnach:
Addresses https://github.com/appium/python-client/issues/878

### DIFF
--- a/appium/webdriver/webdriver.py
+++ b/appium/webdriver/webdriver.py
@@ -211,6 +211,7 @@ class WebDriver(
         desired_capabilities: Optional[Dict] = None,
         # TODO: Remove the deprecated arg
         browser_profile: Union[str, None] = None,
+        # TODO: Remove the deprecated arg
         proxy: Union[str, None] = None,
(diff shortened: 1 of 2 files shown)


appium/webdriver/webdriver.py
```diff
@@ -232,12 +232,23 @@ def __init__(
...
+            AppiumOptions().load_capabilities(desired_capabilities)
+            if desired_capabilities is not None and options is None
+            else options
+        )
+
         super().__init__(
             command_executor=command_executor,
-            desired_capabilities=desired_capabilities,
             browser_profile=browser_profile,
             proxy=proxy,
```
KazuCocoa, line 255: Maybe we should remove browser_profile and proxy as well?
https://github.com/SeleniumHQ/selenium/pull/12030/files#diff-c87c150ed2a08f5293db58a06da8bf7f14c3a5582586c6270c1be661dfdd0985L205

mykola-mokhnach: done

PR #889, opened 2023-07-07
chore: update isort revision to 5.12.0

Dor-bl:
using `isort` revision 5.10.1 raise an error when trying to commit a change:
```
line 58, in create_poetry
          raise RuntimeError("The Poetry configuration is invalid:\n" + message)
      RuntimeError: The Poetry configuration is invalid:
        - [extras.pipfile_deprecated_finder.2] 'pip-shims<=0.3.4' does not match '^[a-zA-Z-_.0-9]+$'

      [end of output]
```
once I updated `isort` to 5.12.0, the error is gone.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/PyCQA/isort
-    rev: 5.10.1
+    rev: 5.12.0
     hooks:
       - id: isort
         args: [ "." ]


.pre-commit-config.yaml
```diff
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/PyCQA/isort
-    rev: 5.10.1
+    rev: 5.12.0
```
KazuCocoa, line 3: Let me check if we can remov this line since we basically manage versions in https://github.com/appium/python-client/blob/b142e00e3f47065752c80f71f2f8cb80bf2500f7/Pipfile#L9 . I guess this version can be removed. (but it is not this pr)

KazuCocoa (approved):

PR #890, opened 2023-07-08
chore: run pre-commit autoupdate

KazuCocoa:
Ran `pre-commit autoupdate`. Pipfile.lock has the same version.

I wondered if we could sync the rev and Pipfile.lock, but probably no good way...?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,13 +5,13 @@ repos:
       - id: isort
         args: [ "." ]
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.961
+    rev: v1.4.1
     hooks:
       - id: mypy
         entry: mypy appium/ test/functional
         pass_filenames: false
   - repo: https://github.com/psf/black
-    rev: 22.3.0
+    rev: 23.3.0
     hooks:
       - id: black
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -5,13 +5,13 @@ repos:
       - id: isort
         args: [ "." ]
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.961
+    rev: v1.4.1
```
mykola-mokhnach, line 8: would it be possible to just use the most recent snapshot there instead of a fixed revision?

KazuCocoa: Initially wondered if the rev could be removed. Let me research further later.

I saw the rev was optional in a stack overflow but the official doc didn't address that.

Dor-bl: According to GPT, we can use: `rev: latest`
didn't tried it yet though :)

KazuCocoa: it seems like `pre-commit autoupdate` is the only thing we could do to update versions, so maybe the best way is run `pre-commit autoupdate` via CI task and commit it on it...?

Anyway, this pr has nothing we can do more

mykola-mokhnach (approved):

PR #909, opened 2023-09-01
refactor!: remove several previously deprecated APIs

mykola-mokhnach:
BREAKING CHANGE: Removed obsolete all_sessions and session properties
BREAKING CHANGE: Removed the obsolete start_activity method
BREAKING CHANGE: Removed the obsolete end_test_coverage method
BREAKING CHANGE: Removed the following obsolete arguments from the driver constructor: desired_capabilities, browser_profile, proxy
BREAKING CHANGE: Removed obsolete set_value and set_text methods
BREAKING CHANGE: Removed the obsolete MobileBy class
BREAKING CHANGE: Removed obsolete application management methods: launch_app, close_app, reset
BREAKING CHANGE: Removed obsolete IME methods: available_ime_engines, is_ime_active, activate_ime_engine, deactivate_ime_engine, active_ime_engine

PR #919, opened 2023-10-01
Add back `set_value` and remove deprecated mark

easyteacher:
Otherwise there is no way to change a slider's value in Appium tests.

This partially reverts https://github.com/appium/python-client/pull/909

Ref: https://invent.kde.org/sdk/selenium-webdriver-at-spi/-/blob/f17f6fbbe7175e96ceb89ea3426901875d20228a/selenium-webdriver-at-spi.py#L735

### DIFF
--- a/appium/webdriver/mobilecommand.py
+++ b/appium/webdriver/mobilecommand.py
@@ -33,6 +33,8 @@ class MobileCommand:
     TOUCH_ACTION = 'touchAction'
     MULTI_ACTION = 'multiAction'
 
+    SET_IMMEDIATE_VALUE = 'setImmediateValue'
+
     BACKGROUND = 'background'
     GET_APP_STRINGS = 'getAppStrings'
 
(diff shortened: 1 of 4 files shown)


KazuCocoa:
Have you tried `element.send_keys`? The endpoint to UIA2 server is https://github.com/appium/appium-uiautomator2-driver/blob/b3dd0a2738543a0ce3ceeb6d3dfe8aa2918977d4/lib/commands/element.js#L68-L74 that is the same as https://github.com/appium/appium/blob/master/packages/base-driver/lib/protocol/routes.js#L234-L248 (element.send_keys)

easyteacher:
Looks like the official webdriver does not really use set_value. But in  selenium-webdriver-at-spi send_keys and set_values map to "Send key input" and "atspi.valueElement.currentValue = value", so they are different.

mykola-mokhnach:
The idea is to only have send_keys.

You can still have non-standard api if it is used with your particular driver by adding an extension method to the driver. Check https://github.com/appium/python-client/blob/fa7ba6ee5ee91c914f69e34547993af62995462f/test/unit/webdriver/webdriver_test.py#L210 as an example

easyteacher:
Nice, didn't know that. Thanks for pointing out.

PR #934, opened 2023-11-28
feat: support list attribute

kkb912002:


KazuCocoa:
Could you add test code like https://github.com/appium/python-client/blob/876233e115ef68839f614e89f3dd7bd523222d36/test/unit/webdriver/webelement_test.py#L66 ?
Please sign the CLA as well.

appium/webdriver/webelement.py
```diff
@@ -60,7 +60,7 @@ def get_attribute(self, name: str) -> Optional[Union[str, Dict]]:
         if attribute_value is None:
             return None
 
-        if isinstance(attribute_value, dict):
+        if isinstance(attribute_value, dict) or isinstance(attribute_value, list):
```
mykola-mokhnach, line 63: I assume the return type must be updated as well and then (probably) mypy output fixed

mykola-mokhnach: also isinstance may accept a tuple as the second argument

mykola-mokhnach:
Could you please provide some particular examples of where list is returned as attribute value? In java client all attributes are converted to strings: https://github.com/SeleniumHQ/selenium/blob/36585d189b2e9f2ced136a7e6c456ffe53604141/java/src/org/openqa/selenium/remote/RemoteWebElement.java#L149

kkb912002:
@mykola-mokhnach  
If the 'value' field is parseable as JSON, it might be automatically parsed by the HTTP library used by the client in each language.
If there is a standard, we should follow it. This change follows the context of returning the dictionary as is. If it needs to be returned as a string, using `json.dumps` would be necessary for dict or list in the current situation.

ex) original value(string): '["abc", "def"]'
Parsed as List => ['abc', 'def']
- str() => "['abc', 'def']" (value changed can't be parsed as json)
- json.dumps() => '["abc", "def"]'

What should we do?

mykola-mokhnach:
> If there is a standard, we should follow it. This change follows the context of returning the dictionary as is. If it needs to be returned as a string, using json.dumps would be necessary for dict or list in the current situation.

W3C Spec needs the result to be a string. I don't know why somebody added dict there as a special return case. I assume it would be smart to perform `json.dumps` on list values for this particular case without changing the return type. I would also do the same for dictionaries, but this will be a breaking change, so maybe in the next major client update

KazuCocoa:
@kkb912002 Could you give us what driver and element caused the case? Perhaps we could check the behavior as well

kkb912002:
> @kkb912002 Could you give us what driver and element caused the case? Perhaps we could check the behavior as well

This is specific to occurrences within my custom UI library. I plan to make changes to comply with the W3C spec, returning a string. Feel free to close this request.

PR #940, opened 2023-12-13
fix: self.command_executor instance in _update_command_executor

KazuCocoa:
The new replaced instance was always `RemoteConnection`, which was not Appium's one. Use AppiumConnection if the executor was already made by AppiumConnection

### DIFF
--- a/appium/webdriver/webdriver.py
+++ b/appium/webdriver/webdriver.py
@@ -296,8 +296,7 @@ class WebDriver(
 
         logger.debug('Updated request endpoint to %s', executor)
         # Override command executor.
-        # TODO: add custome executor case by the user.
-        if isinstance(self.command_executor, AppiumConnection):
+        if isinstance(self.command_executor, AppiumConnection):  # type: ignore
             self.command_executor = AppiumConnection(executor, keep_alive=keep_alive)
         else:
             self.command_executor = RemoteConnection(executor, keep_alive=keep_alive)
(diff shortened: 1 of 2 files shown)


appium/webdriver/webdriver.py
```diff
@@ -295,8 +295,11 @@ def _update_command_executor(self, keep_alive: bool) -> None:
         executor = f'{protocol}://{hostname}:{port}{path}'
 
         logger.debug('Updated request endpoint to %s', executor)
-        # Override command executor
-        self.command_executor = RemoteConnection(executor, keep_alive=keep_alive)
+        # Override command executor.
+        if isinstance(self.command_executor, AppiumConnection):  # type: ignore
```
mykola-mokhnach, line 299: why should we ignore mypy here?

KazuCocoa: I need to take a look later as a separate PR (so minimize each file change). For now, the below error kept rising without `ignore`. Maybe it needs a couple of lines' update with type hints.

```
appium/webdriver/webdriver.py:300: error: Cannot determine type of "command_executor"  [has-type]
```

mykola-mokhnach: this looks like a mypy bug.
A similar issue has been reported here: https://github.com/python/mypy/issues/13700

Lets keep it like that for now

mykola-mokhnach (approved):

PR #944, opened 2023-12-24
docs: update links

KazuCocoa:
http://appium.io/docs/en/commands/interactions/actions/ returns 404.

selenium documentation has actions api right now. This lib's README has `MultiAction/TouchAction to W3C actions` thus at least we can replace the 404 link to the README.

cc @Dor-bl

### DIFF
--- a/appium/webdriver/common/multi_action.py
+++ b/appium/webdriver/common/multi_action.py
@@ -33,7 +33,7 @@ if TYPE_CHECKING:
 class MultiAction:
     """
     deprecated:: 2.0.0
-        Please use W3C actions instead: http://appium.io/docs/en/commands/interactions/actions/
+        Please use W3C actions instead: https://github.com/appium/python-client?tab=readme-ov-file#multiactiontouchaction-to-w3c-actions
     """
 
     def __init__(self, driver: 'WebDriver', element: Optional['WebElement'] = None) -> None:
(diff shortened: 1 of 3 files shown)


appium/webdriver/common/multi_action.py
```diff
@@ -33,7 +33,7 @@
 class MultiAction:
     """
     deprecated:: 2.0.0
-        Please use W3C actions instead: http://appium.io/docs/en/commands/interactions/actions/
+        Please use W3C actions instead: https://github.com/appium/python-client?tab=readme-ov-file#multiactiontouchaction-to-w3c-actions
```
Dor-bl, line 36: Or 'mobile:' extensions?

appium/webdriver/common/touch_action.py
```diff
@@ -37,7 +37,7 @@
 class TouchAction:
     """
     deprecated:: 2.0.0
-        Please use W3C actions instead: http://appium.io/docs/en/commands/interactions/actions/
+        Please use W3C actions instead: https://github.com/appium/python-client?tab=readme-ov-file#multiactiontouchaction-to-w3c-actions
```
Dor-bl, line 40: Or 'mobile:'  extensions?

appium/webdriver/common/multi_action.py
KazuCocoa: Actual replacement is W3C actions for the touch actions APIs. The mobile extension is possible alternative (could be not equivalent behavior) thus I think actions api is sufficient as API documentation.
README includes the alternative as well though as part of migration guide.

Dor-bl (approved):

PR #945, opened 2024-01-09
docs: update example in readme

KazuCocoa:
@Dor-bl Was this readme thing in the slack's question?


Dor-bl:
> @Dor-bl Was this readme thing in the slack's question?

Nope. It was not.
Thanks for the update

PR #987, opened 2024-05-06
ci: use gha instead of Azure for iOS in Azure

KazuCocoa:
functional-test.yml can optimize more, but for now, this PR aims to drop Azure tests for iOS.
I'll update functional-test.yml for GHA after adding Android to remove azure.


.github/workflows/functional-test.yml
```diff
@@ -43,21 +47,29 @@ jobs:
         model: 'iPhone 15 Plus'
         os_version: '17.4'
```
mykola-mokhnach, line 48: It makes sense to extract these values and keep them in a single place, for example  env variables
The reason for it is that when we need to change iOS version this also usually requires to change phone model and macOS version

test/functional/ios/helper/desired_capabilities.py
```diff
@@ -28,7 +28,7 @@ def get_desired_capabilities(app: Optional[str] = None) -> Dict[str, Any]:
     desired_caps: Dict[str, Any] = {
         'deviceName': iphone_device_name(),
         'platformName': 'iOS',
-        'platformVersion': '15.5',
+        'platformVersion': '17.4',
```
mykola-mokhnach, line 31: this value should be received from CI env

test/functional/ios/helper/desired_capabilities.py
```diff
@@ -28,7 +28,7 @@ def get_desired_capabilities(app: Optional[str] = None) -> Dict[str, Any]:
     desired_caps: Dict[str, Any] = {
         'deviceName': iphone_device_name(),
```
mykola-mokhnach, line 29: this value should be received from CI env

test/functional/ios/helper/desired_capabilities.py
```diff
@@ -70,7 +70,7 @@ def wda_port() -> int:
 
 
 def iphone_device_name() -> str:
-    prefix = 'iPhone 12'
+    prefix = 'iPhone 15 Plus'
```
mykola-mokhnach, line 73: this value should be received from CI env

test/unit/helper/test_helper.py
```diff
@@ -98,7 +98,7 @@ def ios_w3c_driver() -> 'WebDriver':
             'sessionId': '1234567890',
             'capabilities': {
                 'device': 'iphone',
-                'browserName': 'UICatalog',
+                'browserName': 'UIKitCatalog',
```
mykola-mokhnach, line 101: should this value be moved into constants together with the above ones?

.github/workflows/functional-test.yml
KazuCocoa: Yes, similar thing will occur in Android as well, I believe. So firsly I move Azure to GHA without such definition and will do for Android and iOS both later. Then perhaps it would make sense to modify python code side as well to refer to them.

test/functional/ios/search_context/find_by_element_webelement_tests.py
```diff
@@ -18,14 +18,14 @@
 
 class TestFindByElementWebelement(BaseTestCase):
     def test_find_element_by_path(self) -> None:
-        el = self.driver.find_element(by=AppiumBy.IOS_PREDICATE, value='wdName == "UICatalog"')
-        assert 'UICatalog' == el.get_attribute('name')
+        el = self.driver.find_element(by=AppiumBy.IOS_PREDICATE, value='wdName == "UIKitCatalog"')
```
mykola-mokhnach, line 21: should the constant also be used in the located? same below

mykola-mokhnach (approved):

PR #1007, opened 2024-07-23
ci: move Azure to GHA (Android)

KazuCocoa:
Let's run one first test on GHA from Azure.

This includes only  `test/functional/android/device_time_tests.py test/functional/android/search_context/find_by_*.py`


.github/workflows/functional-test.yml
```diff
@@ -27,6 +27,11 @@ jobs:
 
     runs-on: macos-14
 
+    env:
+      XCODE_VERSION: 15.3
```
mykola-mokhnach, line 31: maybe add a note that in order to change xcode version one must also check for compatible macOS version in the compat matrix

.github/workflows/functional-test.yml
```diff
@@ -38,14 +43,14 @@ jobs:
...
-        xcode-version: 15.3
+        xcode-version: ${{ env.XCODE_VERSION }}
     - run: defaults write com.apple.iphonesimulator PasteboardAutomaticSync -bool false
 
     - uses: futureware-tech/simulator-action@v3
       with:
         # https://github.com/actions/runner-images/blob/main/images/macos/macos-14-arm64-Readme.md
-        model: 'iPhone 15 Plus'
-        os_version: '17.4'
+        model: ${{ env.IPHONE_MODEL }}
```
mykola-mokhnach, line 52: should these values be quoted?

.github/workflows/functional-test.yml
```diff
@@ -77,3 +82,101 @@ jobs:
...
+        target: google_apis
+        force-avd-creation: false
+        emulator-options: -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
+        disable-animations: false
+        script: echo "Generated AVD snapshot for caching."
+
+    - name: Set up Python 3.9
+      uses: actions/setup-python@v3
+      with:
+        python-version: 3.9
```
mykola-mokhnach, line 152: should it rather be latest python?

.github/workflows/functional-test.yml
```diff
@@ -77,3 +82,101 @@ jobs:
...
+        node-version: 'lts/*'
+
+    # Start Appium
+    - run: npm install -g appium
+    - run: |
+        appium driver install uiautomator2
+        appium driver install espresso
+        appium plugin install images
+        appium plugin install execute-driver
+        nohup appium --use-plugins=images,execute-driver --relaxed-security --log-timestamp --log-no-colors > appium.log &
```
mykola-mokhnach, line 121: we should also redirect the stderr output to stdout

.github/workflows/functional-test.yml
KazuCocoa: Maybe either is fine? This works right now. I'm good either way.

mykola-mokhnach: fine with me if it works this way

.github/workflows/functional-test.yml
mykola-mokhnach: np, just leave a TODO there for later so we don't forget about it.

mykola-mokhnach (approved):

PR #1008, opened 2024-07-29
ci: run other android tests on GHA

KazuCocoa:


KazuCocoa:
just adding more cases on GHA

Issue #1027, opened 2024-09-23
add possibility to increase stop service timeout


LykashevichAndrei:
appium/webdriver/appium_service.py

please make  timeout for self._process.communicate(timeout=5)

```
def stop(self, timeout=5) -> bool:
        """Stops Appium service if it is running.

        The call will be ignored if the service is not running
        or has been already stopped.

        Returns:
            `True` if the service was running before being stopped
        """
        is_terminated = False
        if self.is_running:
            assert self._process
            self._process.terminate()
            self._process.communicate(timeout=timeout)
            is_terminated = True
        self._process = None
        self._cmd = None
        return is_terminated
```
        
 as a method param

KazuCocoa:
And could you share the timeout case's Appium log? I'm curious about what situation caused over 5 sec to stop the process.

LykashevichAndrei:
2024-09-23 12:49:48:719 [HTTP] Closing Appium HTTP server
2024-09-23 12:49:53:722 [HTTP] Not all active connections have been closed within 5001ms. Exiting anyway.

mac m3 48gb 
macos 16 sonoma

appium -v 2.11.4
mac2@1.19.1

speaking about PR - for sure i can provide it.
but it pretty minor fix and i think that someone who already developing some features for python client can add this

LykashevichAndrei:
https://github.com/appium/python-client/pull/1030

testerxiaodong:
When I use the readme examples, I get the error below

@pytest.fixture(scope='session')
    def appium_service():
        service = AppiumService()
        service.start(
            # Check the output of `appium server --help` for the complete list of
            # server command line arguments
            args=['--address', APPIUM_HOST, '-p', str(APPIUM_PORT)],
            timeout_ms=20000,
        )
        yield service
        if service.is_running:
>           service.stop()

conftest.py:23: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../Library/Caches/pypoetry/virtualenvs/appium-practice-u-JbVLem-py3.12/lib/python3.12/site-packages/appium/webdriver/appium_service.py:238: in stop
    self._process.communicate(timeout=5)
../../.pyenv/versions/3.12.0/lib/python3.12/subprocess.py:1209: in communicate
    stdout, stderr = self._communicate(input, endtime, timeout)
[...]

PR #1030, opened 2024-09-24
feat: add timeout for server stop as method param

LykashevichAndrei:


mykola-mokhnach:
I am not sure what you are trying to achieve by this PR. Active connections close timeout is currently [hardcoded](https://github.com/appium/appium/blob/6a8f26a9cbf01a7c9181c90bb999d332a343d94d/packages/base-driver/lib/express/server.js#L34) on the server side and won't be affected by the process timeout.

LykashevichAndrei:
https://github.com/appium/python-client/issues/1027

at least for now i will not see error on test postconditions that appium server was not closed

however this is a good catch and i will review appium server code and may be provide another PR to avoid this hardcode

appium/webdriver/appium_service.py
```diff
@@ -222,20 +222,23 @@ def start(self, **kwargs: Any) -> sp.Popen:
             raise AppiumServiceError(error_msg)
         return self._process
 
-    def stop(self) -> bool:
+    def stop(self, timeout: int = 5) -> bool:
```
mykola-mokhnach, line 225: python timedeltas are represented as float seconds, not integers

mykola-mokhnach:
Please rename the PR to comply with the https://www.conventionalcommits.org/en/v1.0.0/ standard

appium/webdriver/appium_service.py
KazuCocoa, line 225:
```suggestion
    def stop(self, timeout: float = 5) -> bool:
```

mykola-mokhnach:
Closed in favour of https://github.com/appium/python-client/pull/1038

PR #1038, opened 2024-10-03
feat: Add a separate function for service startup validation

mykola-mokhnach:
Also sorted module-private methods and aligned service url parsing logic

### DIFF
--- a/appium/webdriver/appium_service.py
+++ b/appium/webdriver/appium_service.py
@@ -145,9 +145,7 @@ class AppiumService:
             try:
                 self._process.communicate(timeout=timeout)
             except sp.SubprocessError:
-                import signal
-
-                os.kill(self._process.pid, signal.SIGKILL)
+                self._process.kill()
         self._process = None
         self._cmd = None
         return was_running
--- a/appium/webdriver/appium_service.py
+++ b/appium/webdriver/appium_service.py
@@ -127,7 +127,7 @@ class AppiumService:
             raise AppiumServiceError(error_msg)
         return self._process
 
-    def stop(self, timeout=5.5) -> bool:
+    def stop(self, timeout: float = 5.5) -> bool:
         """Stops Appium service if it is running.
 
         The call will be ignored if the service is not running
--- a/appium/webdriver/appium_service.py
+++ b/appium/webdriver/appium_service.py
@@ -17,7 +17,7 @@ import re
 import subprocess as sp
 import sys
 import time
-from typing import Any, List, Optional, Set
+from typing import Any, List, Optional, Set, Callable
 
 from selenium.webdriver.remote.remote_connection import urllib3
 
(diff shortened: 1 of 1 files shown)


appium/webdriver/appium_service.py
```diff
@@ -215,38 +119,44 @@ def start(self, **kwargs: Any) -> sp.Popen:
...
+            was_running = True
             self._process.terminate()
-            self._process.communicate(timeout=5)
-            is_terminated = True
+            try:
+                self._process.communicate(timeout=timeout)
+            except sp.SubprocessError:
+                import signal
+
+                os.kill(self._process.pid, signal.SIGKILL)
```
KazuCocoa, line 150: Would `self._process.kill()` be better for cross platform?
https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate

mykola-mokhnach: process.kill is just an alias to process.terminate, which is already called above

KazuCocoa: for Win? i though for Linux are macOS are sigkill in `kill` and alias to terminate for Win. So in terms of Win, yea, it's just an alias already called, so it calls  terminal twice thing

KazuCocoa: Is it in `if _mswindows:`?

Screenshot 2024-10-03 at 10 43 08 AM

mykola-mokhnach: I did not pay attention to the `else:` block. Changed to use .kill

appium/webdriver/appium_service.py
```diff
@@ -215,38 +119,42 @@ def start(self, **kwargs: Any) -> sp.Popen:
...
         if self.is_running:
             assert self._process
+            was_running = True
             self._process.terminate()
-            self._process.communicate(timeout=5)
-            is_terminated = True
+            try:
+                self._process.communicate(timeout=timeout)
+            except sp.SubprocessError:
+                self._process.kill()
```
KazuCocoa, line 148: actually, after kill, we need wait/communicate
```suggestion
                self._process.kill()
                _, _ = self._process. communicate(timeout=timeout)
```
(or print the error) to ensure the process state

KazuCocoa (approved):
left process handling comment, otherwise lg entirely

appium/webdriver/appium_service.py
mykola-mokhnach: kill sends SIGKILL, which terminates the process unconditionally.
maybe there is some official link to docs about the best way of cross-platform graceful/forceful process termination ?

KazuCocoa: I haven't found out cross platform thing, but my comment to put communicate comes from https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate 's example code. Actually it is not after terminate's exception though.

On Windows, I explicitly called taskkill below with force option when terminate/kill fails.

```
os.system("taskkill /f /pid {}".format(process.pid))
process.wait(timeout=5)
```

Based on my past Win's experience in work, the taskkill helped to kill process forcefully. The terminate/kill in Python was a graceful one. Probably... this taskkill could be ok to add when we get a new issue though

mykola-mokhnach: it's actually a good idea to use taskkill, I'll add it to the implementation

PR #1039, opened 2024-10-03
chore: Use proper type declarations for methods returning self instances

mykola-mokhnach:


KazuCocoa (approved):
nice, didn't know of the "Self" with the library

Issue #1042, opened 2024-10-08
Use ruff instead of pylint, isort and black


KazuCocoa:
I started getting below on my local. It comes from `pylint_quotes`, but the plugin might have some issues.

```
python -m pylint  --load-plugins pylint_quotes --rcfile .pylintrc appium test --disable=E1136
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/Users/kazu/.asdf/installs/python/3.12.4/lib/python3.12/site-packages/pylint/__main__.py", line 10, in <module>
    pylint.run_pylint()
  File "/Users/kazu/.asdf/installs/python/3.12.4/lib/python3.12/site-packages/pylint/__init__.py", line 34, in run_pylint
    PylintRun(argv or sys.argv[1:])
  File "/Users/kazu/.asdf/installs/python/3.12.4/lib/python3.12/site-packages/pylint/lint/run.py", line 155, in __init__
    linter.load_plugin_modules(self._plugins)
  File "/Users/kazu/.asdf/installs/python/3.12.4/lib/python3.12/site-packages/pylint/lint/pylinter.py", line 379, in load_plugin_modules
    module = astroid.modutils.load_module_from_name(modname)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kazu/.asdf/installs/python/3.12.4/lib/python3.12/site-packages/astroid/modutils.py", line 196, in load_module_from_name
    module = importlib.import_module(dotted_name)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kazu/.asdf/installs/python/3.12.4/lib/python3.12/importlib/__init__.py", line 90, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1331, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 935, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 995, in exec_module
[...]

PR #1043, opened 2024-10-09
chore: use ruff (isort, pylint and pyflakes) instead of individual isort, pylint and black libraries

KazuCocoa:
https://github.com/appium/python-client/issues/1042

Rules:
https://docs.astral.sh/ruff/rules/

Some customizations such as quote are in `.ruff.toml`. Max line length comes from current pylint rule.
pyflakes is default one in ruff. I have enabled isort and pylint.

Some diffs between `black` and default `ruff`.
https://docs.astral.sh/ruff/formatter/black/

I have applied:
```
ruff format .
ruff check --fix .
```

I'm good with this PR format, but if we want to add rules, I'm happy to hear it.

cc @mykola-mokhnach if you have any preferences


.ruff.toml
```diff
@@ -0,0 +1,31 @@
...
+[lint.per-file-ignores]
+"__init__.py" = [
+    # unused-import
+    "F401",
+    # import violations
+    "E402"
+]
+"**/{test,docs}/*" = [
+    # https://docs.astral.sh/ruff/rules/magic-value-comparison/
+    "PLR2004"
```
KazuCocoa, line 23: This is to prevent error like below:
```
test/unit/webdriver/network_test.py:33:45: PLR2004 Magic value used in comparison, consider replacing `2` with a constant variable
   |
31 |             body='{"value": {"wifi": true, "data": false, "airplaneMode": false}}',
32 |         )
33 |         assert driver.network_connection == 2
   |                                             ^ PLR2004
34 | 
35 |     @httpretty.activate
   |
```

Makefile
```diff
@@ -5,22 +5,10 @@ check-all: ## Run all lint checks and unittest
...
-
-.PHONY: pylint
-pylint: ## Run pylint
-    # TODO Remove --disable=E1136 when no errors in py39
-	python -m pylint $(ARGS) --load-plugins pylint_quotes --rcfile .pylintrc appium test --disable=E1136
-
-.PHONY: mypy
-mypy:  ## Run mypy
-	python -m mypy appium test/functional
+.PHONY: ruff
```
mykola-mokhnach, line 8: maybe create two steps: format + lint for local dev and the step `check`, which would be called automatically in the precommit hook

mykola-mokhnach: the idea is that we don't want to rely on tool names, but rather intention purposes

KazuCocoa: makes sense

mykola-mokhnach (approved):
👍

PR #1045, opened 2024-10-12
[do not merge]

KazuCocoa:
as a rough reference of https://github.com/SeleniumHQ/selenium/pull/14587

This needs to modify local deps directly tho.

I'll create a new PR when I integrate with the selenium's one.

---

https://github.com/SeleniumHQ/selenium/pull/13286

PR #1054, opened 2024-10-31
feat: support selenium 4.26+: support ClientConfig and refactoring internal implementation

KazuCocoa:
Python selenium client has two major changes:
- https://github.com/appium/python-client/pull/1045 related
    - Removed `find_element` and `find_elements` override
    - Removed strict_ssl by this library. Refer to the selenium's implementation (code removal)
    - Used `RemoteConnection.user_agent` to define UA
    - Used `RemoteConnection.extra_headers` to define extra headers
    - Used `self.command_executor.add_command` to define methods
- Use ClientConfig for the AppiumConnection
    - https://github.com/SeleniumHQ/selenium/blob/trunk/py/selenium/webdriver/remote/client_config.py#L29
    - Maybe... https://github.com/SeleniumHQ/selenium/blob/trunk/py/selenium/webdriver/remote/client_config.py#L57C34-L57C56 would help...?
- https://github.com/SeleniumHQ/selenium/pull/14578 This PR introduced hardcoded 120 seconds request timeout
    - It breaks Appium's long new session request. https://github.com/SeleniumHQ/selenium/pull/14692 will fix this issue but we should prepare a guidance about how to set timeout since this behavior change could affect users

## test

Current unit tests cover find_element/s, user agent, extra headers and commands.
The functional tests cover session creation, do some actions and deleting the session.

### DIFF
--- a/appium/webdriver/appium_connection.py
+++ b/appium/webdriver/appium_connection.py
@@ -13,9 +13,8 @@
 # limitations under the License.
 
 import uuid
-from typing import TYPE_CHECKING, Any, Dict, Optional, Union
+from typing import TYPE_CHECKING, Any, Dict, Optional
 
-import urllib3
 from selenium.webdriver.remote.remote_connection import RemoteConnection
 
 from appium.common.helper import library_version
@@ -30,51 +29,16 @@ PREFIX_HEADER = 'appium/'
 class AppiumConnection(RemoteConnection):
     _proxy_url: Optional[str]
 
-    def __init__(
-        self,
-        remote_server_addr: str,
-        keep_alive: bool = False,
-        ignore_proxy: Optional[bool] = False,
-        init_args_for_pool_manager: Union[Dict[str, Any], None] = None,
-    ):
-        # Need to call before super().__init__ in order to pass arguments for the pool manager in the super.
-        self._init_args_for_pool_manager = init_args_for_pool_manager or {}
-
-        super().__init__(remote_server_addr, keep_alive=keep_alive, ignore_proxy=ignore_proxy)
-
-    def _get_connection_manager(self) -> Union[urllib3.PoolManager, urllib3.ProxyManager]:
-        # https://github.com/SeleniumHQ/selenium/blob/0e0194b0e52a34e7df4b841f1ed74506beea5c3e/py/selenium/webdriver/remote/remote_connection.py#L134
-        pool_manager_init_args = {'timeout': self.get_timeout()}
-
-        if self._ca_certs:
-            pool_manager_init_args['cert_reqs'] = 'CERT_REQUIRED'
-            pool_manager_init_args['ca_certs'] = self._ca_certs
-        else:
-            # This line is necessary to disable certificate verification
-            pool_manager_init_args['cert_reqs'] = 'CERT_NONE'
-
-        pool_manager_init_args.update(self._init_args_for_pool_manager)
-
-        if self._proxy_url:
-            if self._proxy_url.lower().startswith('sock'):
-                from urllib3.contrib.socks import SOCKSProxyManager
-
-                return SOCKSProxyManager(self._proxy_url, **pool_manager_init_args)
-            if self._identify_http_proxy_auth():
-                self._proxy_url, self._basic_proxy_auth = self._separate_http_proxy_auth()
-                pool_manager_init_args['proxy_headers'] = urllib3.make_headers(proxy_basic_auth=self._basic_proxy_auth)
-            return urllib3.ProxyManager(self._proxy_url, **pool_manager_init_args)
-
-        return urllib3.PoolManager(**pool_manager_init_args)
+    RemoteConnection.user_agent = f'{PREFIX_HEADER}{library_version()} ({RemoteConnection.user_agent})'
 
     @classmethod
     def get_remote_connection_headers(cls, parsed_url: 'ParseResult', keep_alive: bool = True) -> Dict[str, Any]:
         """Override get_remote_connection_headers in RemoteConnection"""
         headers = RemoteConnection.get_remote_connection_headers(parsed_url, keep_alive=keep_alive)
-        # e.g. appium/0.49 (selenium/3.141.0 (python linux))
-        headers['User-Agent'] = f'{PREFIX_HEADER}{library_version()} ({headers["User-Agent"]})'
         if parsed_url.path.endswith('/session'):
             # https://github.com/appium/appium-base-driver/pull/400
-            headers['X-Idempotency-Key'] = str(uuid.uuid4())
+            RemoteConnection.extra_headers = {'X-Idempotency-Key': str(uuid.uuid4())}
+        else:
+            RemoteConnection.extra_headers = {}
 
         return headers
--- a/appium/webdriver/appium_connection.py
+++ b/appium/webdriver/appium_connection.py
@@ -15,6 +15,7 @@
 import uuid
 from typing import TYPE_CHECKING, Any, Dict, Optional
 
+from selenium.webdriver.remote.client_config import ClientConfig
 from selenium.webdriver.remote.remote_connection import RemoteConnection
 
 from appium.common.helper import library_version
@@ -30,6 +31,21 @@ class AppiumConnection(RemoteConnection):
     _proxy_url: Optional[str]
 
     RemoteConnection.user_agent = f'{PREFIX_HEADER}{library_version()} ({RemoteConnection.user_agent})'
+
+    def __init__(
+        self,
+        remote_server_addr: Optional[str] = None,
+        keep_alive: Optional[bool] = True,
+        init_args_for_pool_manager: Optional[dict] = None,
+        client_config: Optional[ClientConfig] = None,
+    ):
+        if client_config is None:
+            client_config = ClientConfig(remote_server_addr=remote_server_addr)
+        client_config.keep_alive = keep_alive
+        if init_args_for_pool_manager is not None:
+            client_config.init_args_for_pool_manager = init_args_for_pool_manager
+
+        super().__init__(client_config=client_config)
 
     @classmethod
     def get_remote_connection_headers(cls, parsed_url: 'ParseResult', keep_alive: bool = True) -> Dict[str, Any]:
--- a/appium/webdriver/appium_connection.py
+++ b/appium/webdriver/appium_connection.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 import uuid
-from typing import TYPE_CHECKING, Any, Dict, Optional
+from typing import TYPE_CHECKING, Any, Dict, Optional, Union
 
 from selenium.webdriver.remote.client_config import ClientConfig
 from selenium.webdriver.remote.remote_connection import RemoteConnection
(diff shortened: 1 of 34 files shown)


appium/webdriver/appium_connection.py
```diff
@@ -30,51 +30,39 @@
...
-        if self._ca_certs:
-            pool_manager_init_args['cert_reqs'] = 'CERT_REQUIRED'
-            pool_manager_init_args['ca_certs'] = self._ca_certs
-        else:
-            # This line is necessary to disable certificate verification
-            pool_manager_init_args['cert_reqs'] = 'CERT_NONE'
+        if client_config is None:
+            client_config = ClientConfig(remote_server_addr=remote_server_addr)
+        if keep_alive is not None:
+            client_config.keep_alive = keep_alive
```
mykola-mokhnach, line 46: I am not quite sure if we want to mutate the passed client config. At least, this must be explicitly stated in the docs or in the comment

appium/webdriver/appium_connection.py
```diff
@@ -30,51 +30,39 @@
...
-        # e.g. appium/0.49 (selenium/3.141.0 (python linux))
-        headers['User-Agent'] = f'{PREFIX_HEADER}{library_version()} ({headers["User-Agent"]})'
         if parsed_url.path.endswith('/session'):
             # https://github.com/appium/appium-base-driver/pull/400
-            headers['X-Idempotency-Key'] = str(uuid.uuid4())
+            RemoteConnection.extra_headers = {'X-Idempotency-Key': str(uuid.uuid4())}
+        else:
+            RemoteConnection.extra_headers = {}
 
         return headers
```
mykola-mokhnach, line 64: the returned `headers` object won't be the same as it was before. is it used somewhere?

appium/webdriver/webdriver.py
```diff
@@ -62,6 +63,11 @@
 from .webelement import WebElement as MobileWebElement
 
 
+class AppiumLocatorConverter:
+    def convert(self, by, value):
```
mykola-mokhnach, line 67: Please add explicit type definitions

appium/webdriver/appium_connection.py
```diff
@@ -30,51 +30,39 @@
 class AppiumConnection(RemoteConnection):
     _proxy_url: Optional[str]
```
mykola-mokhnach, line 31: is this property still used?

appium/webdriver/appium_connection.py
```diff
@@ -30,51 +30,39 @@
...
         if parsed_url.path.endswith('/session'):
             # https://github.com/appium/appium-base-driver/pull/400
-            headers['X-Idempotency-Key'] = str(uuid.uuid4())
+            RemoteConnection.extra_headers = {'X-Idempotency-Key': str(uuid.uuid4())}
+        else:
+            RemoteConnection.extra_headers = {}
 
         return headers
+
+    # TODO: remove after https://github.com/SeleniumHQ/selenium/pull/14692 merge
```
KazuCocoa, line 66: I'll suspend this PR until this issue will be fixed in selenium. Instead, ill create a new patch version with selenium’s upper limit to not use 4.26

VietND96: This PR is merged, and we will release patch 4.26.1

appium/webdriver/appium_connection.py
KazuCocoa: Not for us, so can be removed.

appium/webdriver/appium_connection.py
```diff
@@ -26,55 +25,33 @@
 
 PREFIX_HEADER = 'appium/'
 
+_HEADER_IDEMOTENCY_KEY = 'X-Idempotency-Key'
```
mykola-mokhnach, line 28: why the constant should be private?

appium/webdriver/webdriver.py
```diff
@@ -200,32 +213,34 @@ class WebDriver(
     Sms,
     SystemBars,
 ):
-    def __init__(
+    def __init__(  # noqa: PLR0913
```
mykola-mokhnach, line 216: is it possible to extract the custom pre-init code into a class or static helper method, so we don't need to ignore the linter warning?

appium/webdriver/appium_connection.py
```diff
@@ -26,55 +25,33 @@
...
-        headers['User-Agent'] = f'{PREFIX_HEADER}{library_version()} ({headers["User-Agent"]})'
+        """Override get_remote_connection_headers in RemoteConnection to control the extra headers.
+        This method will be used in sending a request method in this class.
+        """
+
         if parsed_url.path.endswith('/session'):
             # https://github.com/appium/appium-base-driver/pull/400
-            headers['X-Idempotency-Key'] = str(uuid.uuid4())
+            cls.extra_headers[_HEADER_IDEMOTENCY_KEY] = str(uuid.uuid4())
+        elif _HEADER_IDEMOTENCY_KEY in cls.extra_headers:
```
mykola-mokhnach, line 54: we must make sure this comparison is case-insensitive to avoid possible issues in the future

appium/webdriver/webdriver.py
KazuCocoa: Do you mean something like:

```
def __init__(self, **kwargs):
    args = getSomething(**kwargs)
    command_executor = args.command_executor
    ...
```
?

I was thinking of defining a custom client config after this PR to drop existing ones such as keep_alive, direct_connect and strict_ssl like `AppiumClientConfig` and:
```
def __init__(
  self,
  command_executor = 'http://127.0.0.1:4444/wd/hub',
  extensions = None,
  client_config = AppiumClientConfig
)
```

with major version update (as a breaking change.) So this lint ignore is a temporary one.

appium/webdriver/appium_connection.py
mykola-mokhnach: yes, I don't see any harm when people could import this header name

appium/webdriver/appium_connection.py
```diff
@@ -26,55 +25,46 @@
...
-            pool_manager_init_args['ca_certs'] = self._ca_certs
-        else:
-            # This line is necessary to disable certificate verification
-            pool_manager_init_args['cert_reqs'] = 'CERT_NONE'
 
-        pool_manager_init_args.update(self._init_args_for_pool_manager)
+def _get_new_headers(key: str, headers: Dict[str, str]) -> Dict[str, str]:
+    """Return a new dictionary of heafers without the given key.
+    The key match is case-insensitive."""
+    new_headers = dict()
```
mykola-mokhnach, line 34: you may use a dictionary comprehension for this purpose

appium/webdriver/webdriver.py
mykola-mokhnach: the original linter error is about super init method is not the first call in the class constructor. In order to address that we would need to move the preceeding code into separate method call(s). for example:

```
def __init__(self, x, y):
    z = ... # lint warning here
    super().__init__(z=z)
```

could be changed to


```
def __init__(self, x, y):
    super().__init__(z=self.calculate_z(x, y)) # no warning

@statsicmethod
def calculate_z(x, y):
    return z
```

KazuCocoa: oh, i see

appium/webdriver/appium_connection.py
KazuCocoa: Sure, I don't have strong opinion to make this private/public. Let me update

appium/webdriver/webdriver.py
KazuCocoa: hm, the error I suppressed was:
```
appium/webdriver/webdriver.py:216:9: PLR0913 Too many arguments in function definition (7 > 6)
    |
214 |     SystemBars,
215 | ):
216 |     def __init__(
    |         ^^^^^^^^ PLR0913
217 |         self,
218 |         command_executor: Union[str, AppiumConnection] = 'http://127.0.0.1:4444/wd/hub',
    |

Found 1 error.
```
so the number of args for the init itself...?


I did very roughly as below but this didn't help:
```
def get_config(command_executor, keep_alive, strict_ssl, client_config):
    if isinstance(command_executor, str):
        client_config = client_config or ClientConfig(
            remote_server_addr=command_executor, keep_alive=keep_alive, ignore_certificates=not strict_ssl
        )
        client_config.remote_server_addr = command_executor
        return client_config
    return None

def get_command_executor(command_executor, keep_alive, strict_ssl, client_config):
    if isinstance(command_executor, str):
        client_config = client_config or ClientConfig(
            remote_server_addr=command_executor, keep_alive=keep_alive, ignore_certificates=not strict_ssl
        )
        client_config.remote_server_addr = command_executor
        return AppiumConnection(remote_server_addr=command_executor, client_config=client_config)
    return command_executor

class WebDriver(
    ...
):
    def __init__(  # noqa: PLR0913
        self,
        command_executor: Union[str, AppiumConnection] = 'http://127.0.0.1:4444/wd/hub',
        keep_alive: bool = True,
        direct_connection: bool = True,
        extensions: Optional[List['WebDriver']] = None,
        strict_ssl: bool = True,
        options: Union[AppiumOptions, List[AppiumOptions], None] = None,
        client_config: Optional[ClientConfig] = None,
    ):
        super().__init__(
            command_executor=get_command_executor(command_executor, keep_alive, strict_ssl, client_config),
            options=options,
            locator_converter=AppiumLocatorConverter(),
            web_element_cls=MobileWebElement,
            client_config=get_config(command_executor, keep_alive, strict_ssl, client_config),
        )

Comment (mykola-mokhnach) [reply appium/webdriver/webdriver.py:216]: I thought this warning is related to a different cause. Please keep the `noqa` then. Nevertheless I like the above change. Please do not forget to add type declarations to new functions

File: appium/webdriver/webdriver.py [hunk 10]
```diff
@@ -200,32 +213,34 @@ class WebDriver(
...
+        elif isinstance(command_executor, AppiumConnection) and strict_ssl is False:
+            logger.warning(
+                "Please set 'ignore_certificates' in the given 'appium.webdriver.appium_connection.AppiumConnection' or "
+                "'selenium.webdriver.remote.client_config.ClientConfig' instead. Ignoring."
+            )
 
         super().__init__(
             command_executor=command_executor,
             options=options,
+            locator_converter=AppiumLocatorConverter(),
```
mykola-mokhnach, line 241: does this cache AppiumLocatorConverter instance or a new one is created every time a new WebDriver instance is constructed?

setup.py
```diff
@@ -49,5 +49,5 @@
         'Topic :: Software Development :: Quality Assurance',
         'Topic :: Software Development :: Testing',
     ],
-    install_requires=['selenium ~= 4.12, < 4.26'],
+    install_requires=['selenium ~= 4.26'],
```
mykola-mokhnach, line 52: should it be `'selenium ~= 4.26, < 5.0'` instead?

appium/webdriver/appium_connection.py
```diff
@@ -26,55 +25,40 @@
...
-            pool_manager_init_args['ca_certs'] = self._ca_certs
-        else:
-            # This line is necessary to disable certificate verification
-            pool_manager_init_args['cert_reqs'] = 'CERT_NONE'
 
-        pool_manager_init_args.update(self._init_args_for_pool_manager)
+def _get_new_headers(key: str, headers: Dict[str, str]) -> Dict[str, str]:
+    """Return a new dictionary of heafers without the given key.
+    The key match is case-insensitive."""
+    key = key.lower()
```
mykola-mokhnach, line 34: I usually try to avoid reassigning of method/function arguments

appium/webdriver/webdriver.py
KazuCocoa: Use given one or new instance is created every time.
https://github.com/SeleniumHQ/selenium/blob/6f99c7029876ef69e20b3dee2d6e0e9348ece7f2/py/selenium/webdriver/remote/webdriver.py#L234

setup.py
KazuCocoa: done. sounds good

appium/webdriver/appium_connection.py
KazuCocoa: yea. done

mykola-mokhnach (approved):

PR #1070, opened 2024-11-28
feat: define AppiumClientConfig

KazuCocoa:
BREAKING CHANGE: Some of WebDriver.Remote constructor arguments were moved to AppiumClientConfig:

- direct_connection
- keep_alive
- strict_ssl

https://github.com/SeleniumHQ/selenium/blob/trunk/py/selenium/webdriver/remote/client_config.py

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,6 @@ For example, some changes in the Selenium binding could break the Appium client.
         # after
         from appium.webdriver.client_config import AppiumClientConfig
         client_config = AppiumClientConfig(
-            remote_server_addr=SERVER_URL_BASE,
             direct_connection=True,
             keep_alive=False,
             ignore_certificates=True,
--- a/README.md
+++ b/README.md
@@ -60,8 +60,8 @@ For example, some changes in the Selenium binding could break the Appium client.
 
 
 ### Quick migration guide from v4 to v5
-- This change affecs only for a user who speficies `keep_alive`, `direct_connection` and `strict_ssl` arguments for `webdriver.Remote`:
-    - Please use `AppiumClientConfig` as `client_config` arguemnt as below:
+- This change affects only for users who specify `keep_alive`, `direct_connection` and `strict_ssl` arguments for `webdriver.Remote`:
+    - Please use `AppiumClientConfig` as `client_config` argument similar to how it is specified below:
         ```python
         SERVER_URL_BASE = 'http://127.0.0.1:4723'
         # before
@@ -86,7 +86,8 @@ For example, some changes in the Selenium binding could break the Appium client.
             client_config=client_config
         )
         ```
-        - Note that you can keep using `webdriver.Remote(url, options=options, client_config=client_config)` format as well. Then, the `remote_server_addr` in `AppiumClientConfig` will prior than the `url` specified via `webdriver.Remote`
+        - Note that you can keep using `webdriver.Remote(url, options=options, client_config=client_config)` format as well.
+        In such case the `remote_server_addr` argument of `AppiumClientConfig` constructor would have priority over the `url` argument of `webdriver.Remote` constructor.
 - Use `http://127.0.0.1:4723` as the default server url instead of `http://127.0.0.1:4444/wd/hub`
 
 ### Quick migration guide from v3 to v4
--- a/README.md
+++ b/README.md
@@ -76,12 +76,12 @@ For example, some changes in the Selenium binding could break the Appium client.
         # after
         from appium.webdriver.client_config import AppiumClientConfig
         client_config = AppiumClientConfig(
+            remote_server_addr=SERVER_URL_BASE,
             direct_connection=True,
             keep_alive=False,
             ignore_certificates=True,
         )
         driver = webdriver.Remote(
-            SERVER_URL_BASE,
             options=UiAutomator2Options().load_capabilities(desired_caps),
             client_config=client_config
         )
--- a/README.md
+++ b/README.md
@@ -58,6 +58,35 @@ For example, some changes in the Selenium binding could break the Appium client.
 > to keep compatible version combinations.
 
 
+### Quick migration guide from v4 to v5
+- Please use `AppiumClientConfig` as `client_config` arguemnt in favor of client arguments below
+    - `keep_alive`, `direct_connection` and `strict_ssl` arguments.
+        ```python
+        SERVER_URL_BASE = 'http://127.0.0.1:4723'
+        # before
+        driver = webdriver.Remote(
+            SERVER_URL_BASE,
+            options=UiAutomator2Options().load_capabilities(desired_caps),
+            direct_connection=True,
+            keep_alive=False,
+            strict_ssl=False
+        )
+
+        # after
+        from appium.webdriver.client_config import AppiumClientConfig
+        client_config = AppiumClientConfig(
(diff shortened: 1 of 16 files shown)


testerxiaodong:
When will this PR be merged?

README.md
```diff
@@ -59,6 +59,34 @@ For example, some changes in the Selenium binding could break the Appium client.
 > to keep compatible version combinations.
 
 
+### Quick migration guide from v4 to v5
+- Please use `AppiumClientConfig` as `client_config` arguemnt in favor of client arguments below
```
mykola-mokhnach, line 63: typo `argument`

appium/webdriver/webdriver.py
```diff
@@ -204,28 +205,21 @@ class WebDriver(
...
-        direct_connection: bool = True,
+        command_executor: Union[str, AppiumConnection] = 'http://127.0.0.1:4723',
         extensions: Optional[List['WebDriver']] = None,
-        strict_ssl: bool = True,
+        file_detector: Optional[FileDetector] = None,
         options: Union[AppiumOptions, List[AppiumOptions], None] = None,
-        client_config: Optional[ClientConfig] = None,
+        client_config: Optional[AppiumClientConfig] = None,
     ):
         if isinstance(command_executor, str):
```
mykola-mokhnach, line 214: would it be possible to extract this logic into a separate method, so super class init would be the very first call there?

appium/webdriver/webdriver.py
```diff
@@ -204,28 +205,21 @@ class WebDriver(
 ):
     def __init__(  # noqa: PLR0913
```
mykola-mokhnach, line 206: does it still complain about too many arguments after this change?

appium/webdriver/webdriver.py
```diff
@@ -204,28 +205,21 @@ class WebDriver(
...
         self,
-        command_executor: Union[str, AppiumConnection] = 'http://127.0.0.1:4444/wd/hub',
-        keep_alive: bool = True,
-        direct_connection: bool = True,
+        command_executor: Union[str, AppiumConnection] = 'http://127.0.0.1:4723',
```
mykola-mokhnach, line 208: this is also a breaking change

appium/webdriver/webdriver.py
KazuCocoa: > super class init would be the very first call there?

Actually we could by modifying here like below:
```
    def __init__(
        self,
        command_executor: AppiumConnection,
        client_config: AppiumClientConfig,
        extensions: Optional[List['WebDriver']] = None,
        file_detector: Optional[FileDetector] = None,
        options: Union[AppiumOptions, List[AppiumOptions], None] = None,
    ):
```

Then, a new method will help to build client_config and command_executor.

To avoid handling string of `command_executor` in selenium, which uses `ClientConfig`,  I have added these lines to keep using `AppiumClientConfig` as appium python client for such a string case (to keep providing similar syntax to webdriver.Remote for selenium and appium)

appium/webdriver/webdriver.py
```diff
@@ -174,6 +174,27 @@ def add_command(self) -> Tuple[str, str]:
         raise NotImplementedError()
 
 
+def _get_remote_connection_and_client_config(
+    command_executor: Union[str, AppiumConnection], client_config: Optional[AppiumClientConfig]
```
mykola-mokhnach, line 178: should the optional value be assigned to None by default?

README.md
```diff
@@ -59,6 +59,35 @@ For example, some changes in the Selenium binding could break the Appium client.
...
 
 
+### Quick migration guide from v4 to v5
+- Please use `AppiumClientConfig` as `client_config` argument in favor of client argument below
+    - `keep_alive`, `direct_connection` and `strict_ssl` arguments.
+        ```python
+        SERVER_URL_BASE = 'http://127.0.0.1:4723'
+        # before
+        driver = webdriver.Remote(
+            SERVER_URL_BASE,
```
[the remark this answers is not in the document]
mykola-mokhnach: upd: I can observe the below code supports such feature, so we just need to mention that in documents

appium/webdriver/webdriver.py
```diff
@@ -237,8 +246,8 @@ def __init__(  # noqa: PLR0913
```
mykola-mokhnach, line 246: I believe `if hasattr(self, 'command_executor')` would always return True

README.md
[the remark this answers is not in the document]
KazuCocoa: Sounds good.

The ClinetConfig requires `remote_server_addr` https://github.com/SeleniumHQ/selenium/blob/dbf3daef5519b0a35a8408510854ec7766455c66/py/selenium/webdriver/remote/client_config.py#L80C9-L80C27 so current example is to reduce adding `remote_server_addr` to both, but actually current one will confuse users.

I'll add this.

appium/webdriver/webdriver.py
KazuCocoa: Either is fine. Just wanted to let the caller explicitly set both

mykola-mokhnach (approved):

README.md
```diff
@@ -59,6 +59,36 @@ For example, some changes in the Selenium binding could break the Appium client.
 > to keep compatible version combinations.
 
 
+### Quick migration guide from v4 to v5
+- This change affecs only for a user who speficies `keep_alive`, `direct_connection` and `strict_ssl` arguments for `webdriver.Remote`:
```
mykola-mokhnach, line 63: This change affects only user who specify ...

README.md
```diff
@@ -59,6 +59,36 @@ For example, some changes in the Selenium binding could break the Appium client.
...
+    - Please use `AppiumClientConfig` as `client_config` arguemnt as below:
```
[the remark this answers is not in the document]
mykola-mokhnach: as below -> similar to how it is specified below

README.md
```diff
@@ -59,6 +59,36 @@ For example, some changes in the Selenium binding could break the Appium client.
...
+            direct_connection=True,
+            keep_alive=False,
+            ignore_certificates=True,
+        )
+        driver = webdriver.Remote(
+            options=UiAutomator2Options().load_capabilities(desired_caps),
+            client_config=client_config
+        )
+        ```
+        - Note that you can keep using `webdriver.Remote(url, options=options, client_config=client_config)` format as well. Then, the `remote_server_addr` in `AppiumClientConfig` will prior than the `url` specified via `webdriver.Remote`
```
mykola-mokhnach, line 89: Then, the ... -> In such case the `remote_server_addr` value of `AppiumClientConfig` would have priority over the `url` argument of `webdriver.Remote`

appium/webdriver/webdriver.py
```diff
@@ -174,6 +174,27 @@ def add_command(self) -> Tuple[str, str]:
...
+    if not isinstance(command_executor, str):
+        # client config already defined in the custom command executor
+        # will be prior than the given one.
+        return (command_executor, None)
+
+    # command_executor is str
+    if client_config is None:
+        # Do not keep None to avoid warnings in Selenium
+        # which can prevent with ClientConfig instance usage.
+        client_config = AppiumClientConfig(remote_server_addr=command_executor)
```
mykola-mokhnach, line 194: would it be possible to do the same without reassigning method arguments?

Issue #644, opened 2021-10-15
Update test code for python 3


KazuCocoa:
Current test code inherits `object` and can make them more pytest style.
https://github.com/appium/python-client/pull/636#discussion_r729545712

navin772:
Hi @KazuCocoa, what else needs to be changed apart from removing inheritance of the `object`?

KazuCocoa:
The syntax related is the object. I think we have more minor things, but I don't have concrete examples right now.

PR #1095, opened 2025-02-13
test: Move `app_test.py` to pytest style

Dor-bl:
use pytest without class-based structures, using parameterization for better reusability.

Partly address https://github.com/appium/python-client/issues/644

### DIFF
--- a/test/unit/webdriver/app_test.py
+++ b/test/unit/webdriver/app_test.py
@@ -11,267 +11,154 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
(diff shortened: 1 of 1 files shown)


test/unit/webdriver/app_test.py
```diff
@@ -11,267 +11,154 @@
...
-
 import httpretty
+import pytest
 
 from appium.webdriver.applicationstate import ApplicationState
 from appium.webdriver.webdriver import WebDriver
 from test.unit.helper.test_helper import android_w3c_driver, appium_command, get_httpretty_request_body, ios_w3c_driver
 
 
-class TestWebDriverAppAndroid(object):
```
KazuCocoa, line 22: I initially thought just removing `object` for example, but this is also ok.

PR #1132, opened 2025-06-02
chore: remove unused commands

KazuCocoa:
They have already been removed.

### DIFF
--- a/appium/webdriver/mobilecommand.py
+++ b/appium/webdriver/mobilecommand.py
@@ -30,9 +30,6 @@ class MobileCommand:
     GET_CURRENT_CONTEXT = 'getCurrentContext'
     SWITCH_TO_CONTEXT = 'switchToContext'
 
-    TOUCH_ACTION = 'touchAction'
-    MULTI_ACTION = 'multiAction'
-
     BACKGROUND = 'background'
     GET_APP_STRINGS = 'getAppStrings'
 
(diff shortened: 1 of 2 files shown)


appium/webdriver/mobilecommand.py
```diff
@@ -30,9 +30,6 @@ class MobileCommand:
     GET_CURRENT_CONTEXT = 'getCurrentContext'
     SWITCH_TO_CONTEXT = 'switchToContext'
 
-    TOUCH_ACTION = 'touchAction'
```
mykola-mokhnach, line 33: This is a breaking change. We need to bump the major client version in order to remove deprecated endpoints

KazuCocoa: Do you mean for users who implement this endpoint by themselves by referencing this constant? 
This client's implementation for this has been removed in the Python client v3 -> v4, so these are leftover then.

Then, I'll bump this major version after this merge.

mykola-mokhnach: I did not know the client implementation has been already removed. Then it's probably fine without bumping the major

mykola-mokhnach (approved):

appium/webdriver/mobilecommand.py
KazuCocoa: Ah, gotcha

PR #1155, opened 2025-08-06
feat: Switch package management from pipenv to uv

mykola-mokhnach:


KazuCocoa (approved):
lg, let me merge this and create a release with this for now. Then, will move the push method to `uv`'s one

PR #1160, opened 2025-08-07
chore: use uv for version

KazuCocoa:
`uv version` can be used to manage versions. The getting version code has references, so this PR keeps the version file to return  `importlib.metadata.version('Appium-Python-Client')`. It is available since Python v3.8, and we require Python v3.9+ so should be ok.

### DIFF
--- a/script/release.py
+++ b/script/release.py
@@ -30,7 +30,7 @@ MESSAGE_GREEN = '\033[1;32m{}\033[0m'
 MESSAGE_YELLOW = '\033[1;33m{}\033[0m'
 
 
-def get_current_version():
+def print_current_version():
     os.system('uv version')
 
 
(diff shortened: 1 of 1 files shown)


script/release.py
```diff
@@ -33,9 +31,7 @@
 
 
 def get_current_version():
-    current = io.open(os.path.join(os.path.dirname('__file__'), 'appium', 'version.py'), encoding='utf-8').read().rstrip()
-    print('The current version is {}, type a new one'.format(MESSAGE_YELLOW.format(current)))
-    return current
+    os.system('uv version')
```
mykola-mokhnach, line 34: this method does not return anything, but looks like it should

KazuCocoa: The original one was also to print the version in the stdout, so return value is not matter

mykola-mokhnach: then the method name is confusing as `get` prefix implicitly assumes it must return something

mykola-mokhnach (approved):

Issue #1161, opened 2025-08-07
Use python-semantic-release


KazuCocoa:
- https://github.com/python-semantic-release/python-semantic-release
- https://python-semantic-release.readthedocs.io/en/stable/configuration/configuration-guides/uv_integration.html

KazuCocoa:
Appium Python Client 5.2.0 is now via `uv build` module and the metadata is defined with `pyproject.toml`. Let's see a bit if it will not get any issue report

KazuCocoa:
This includes generating a changelog, so we no longer need `gitchangelog` as well.
https://python-semantic-release.readthedocs.io/en/stable/concepts/getting_started.html#choosing-your-changelog
https://github.com/python-semantic-release/python-semantic-release/blob/5b9d941d5b29da138b933660ce1a9df75b54ce25/CHANGELOG.rst

KazuCocoa:
https://github.com/appium/python-client/pull/1163 is to replace `gitchangelog` with the semantic release lib. The PR is only for changelog stuff.

KazuCocoa:
It looks like GHA does the automatic version determination thing, so the `semantic-release` itself doesn't automatically cound up the semantics version.
https://python-semantic-release.readthedocs.io/en/stable/configuration/automatic-releases/github-actions.html#gh-actions-psr

KazuCocoa:
Automatic release will be done by https://github.com/appium/python-client/issues/1167, but I have dropped the existing release script by replacing them with semantics release to build release stuff and `uv publish` (to push modules to pypi)

KazuCocoa:
https://github.com/appium/python-client/releases/tag/v5.2.2 is the GH release page

PR #1165, opened 2025-08-14
chore: use semantic-release for most of release script execution

KazuCocoa:
Apply https://github.com/appium/python-client/issues/1161 more

With this, the only remaining deps in `release.sh` is publishing modules to pypi. Tag/github release page handlings will be by the semantics release tag


KazuCocoa:
ok, let me run the release steps once to ensure this semantic-release works

Issue #1169, opened 2025-09-09
🐞 bug:[docs][python-client][images] Mismatch: docs expect string, parameter type is bytes AppiumBy.IMAGE / get_images_similarity


wooyoung294:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Appium Version

2.19.0

### Appium Host Type

Command Line

### Is this issue reproducible using the latest components?

- [x] I confirm the issue is still reproducible with the latest component versions

### Current Behavior

**Environment**
python: 3.9
appium-python-client : 5.2.2

**Expected**
(docs): image selector/params should be a base64-encoded string.

**Actual**
(Python client): parameter type is bytes

<img width="991" height="223" alt="Image" src="https://github.com/user-attachments/assets/9c8a22c6-d10c-4953-a83c-666558f14970" />

### Expected Behavior

If this issue is valid, I’d like to submit a PR with the following change to align the parameter type with the docs (string) while still accepting bytes:

```
def get_images_similarity(
    self,
    base64_image1: Union[str, bytes],
    base64_image2: Union[str, bytes],
    **opts: Any
) -> Dict[str, Any]:
    ...
```

### Environment Details

- Debug details (`appium --show-debug-info`):
- Last component version(s) which did _not_ exhibit the problem:
- Platform and version of device under test:
- Real device or emulator/simulator:

### Minimal Reproducible Example
```
from PIL import Image
from appium.webdriver.webdriver import WebDriver
def compare_image_similarity(driver:WebDriver, path, target):
    with Image.open(path) as im:
        buf = io.BytesIO()
        im.save(buf, format="PNG")
        original_png = buf.getvalue()

    target_png = target.screenshot_as_png

    original_png_b64 = base64.b64encode(original_png).decode("ascii")
    target_png_b64 = base64.b64encode(target_png).decode("ascii")

    driver.get_images_similarity(original_png_b64, target_png_b64, visualize=True)
```

mykola-mokhnach:
Thanks for the report. https://github.com/appium/python-client/pull/1170 should fix that

PR #1170, opened 2025-09-09
fix: Adjust images payload for image comparison API

mykola-mokhnach:
Addresses https://github.com/appium/python-client/issues/1169


KazuCocoa (approved):

KazuCocoa:
This change itself looks good. Let me modify the test code side after this merge

PR #1215, opened 2026-04-05
refactor: mobile: getAppStrings to use execute_script exclusively

Dor-bl:
This pull request refactors how app string retrieval is handled in the Appium Python client by removing legacy code paths and cleaning up related commands. The primary focus is on simplifying the `app_strings` method and removing unused or deprecated command definitions.

**App string retrieval refactor:**

* Simplified the `app_strings` method in `applications.py` by removing the fallback to the legacy `GET_APP_STRINGS` command and its associated exception handling. Now, only the extension-based approach is used.

**Command cleanup:**

* Removed the registration of the `GET_APP_STRINGS` command from the command executor in `_add_commands`.
* Deleted the `GET_APP_STRINGS` constant from `MobileCommand` in `mobilecommand.py`, as it is no longer used.


KazuCocoa (approved):

mykola-mokhnach:
typo: fefactor

PR #1269, opened 2026-08-04
feat!: drop CanRememberExtensionPresence

KazuCocoa:
https://github.com/appium/python-client/pull/1267#issuecomment-5166519406


mykola-mokhnach:
wouldn't this be a breaking change?

KazuCocoa:
Yes, this will be. The next release will be a major version update, so I have created this change as well

KazuCocoa:
The test failures for Flutter was the downloaded contents was removed. can be ignored. I'll coordinate it later

KazuCocoa:
I'll prepare a major update with recent removals as well.

PR #1282, opened 2026-08-20
chore: add pre-commit hooks

even-even:
Hello

In PR:
1) add https://github.com/pre-commit/pre-commit-hooks 
2) add pre-commit asserts in makefile


KazuCocoa:
Thank you, but some are duplicated with existing tools. Some are too much to add as precommit for now, so I won't include them in this repo by default.

PR #1283, opened 2026-08-22
chore: add ruff-specific rules

even-even:
Hello

In PR:
1) add https://docs.astral.sh/ruff/rules/#ruff-specific-rules-ruf ruff rule
2) fix warnings in code


KazuCocoa:
I'm good to not add this

PR #1295, opened 2026-09-08
fix: bracket IPv6 hosts in service status URLs

lucasmariano003-wq:
## Description

Appium accepts `--address ::1`, but `AppiumService` builds the status URL without brackets around the IPv6 literal. Its probe therefore uses an invalid URL such as `http://::1:4723/status`, causing startup to fail even when the server is healthy.

Bracket IPv6 hosts when constructing the service URL, without double-bracketing an already bracketed host. IPv4 addresses, hostnames, ports, base paths, argument parsing, and polling behavior are otherwise unchanged.

## Validation

- Reproduced with a real Appium 3.7.0 process on native `::1`, an ephemeral port, and isolated `APPIUM_HOME`. An independent HTTP request confirmed the configured status endpoint was healthy, while the unmodified client's startup check failed on the unbracketed URL.
- After the fix, `AppiumService.start` succeeded over native IPv6. A real IPv4 control also passed. Both server processes were stopped afterward.
- Added five regression cases for loopback/global IPv6 literals, an already bracketed literal, IPv4, and a hostname.
- All 178 unit tests passed; `make check` passed Ruff, formatting, and Mypy over 320 source files.
- Both published files were fetched back and their hashes match the tested copies.

OpenAI Codex performed the implementation and local validation for this GitHub account; maintainer review is pending.

If accepted, please confirm whether this PR qualifies for compensation under the [Appium contributor compensation scheme](https://github.com/appium/appium/blob/master/GOVERNANCE.md#compensation-scheme) and the applicable amount. I understand that classification and payment remain at the project's discretion.

### DIFF
--- a/appium/webdriver/appium_service.py
+++ b/appium/webdriver/appium_service.py
@@ -315,7 +315,10 @@ def _make_status_path(args: list[str]) -> str:
 
 
 def _make_server_url(args: list[str]) -> str:
-    return f'{_parse_protocol(args)}://{_parse_host(args)}:{_parse_port(args)}{_make_status_path(args)}'
+    host = _parse_host(args)
+    if ':' in host and not host.startswith('['):
+        host = f'[{host}]'
+    return f'{_parse_protocol(args)}://{host}:{_parse_port(args)}{_make_status_path(args)}'
 
 
 if __name__ == '__main__':
--- a/appium/webdriver/appium_service.py
+++ b/appium/webdriver/appium_service.py
@@ -18,6 +18,7 @@ import subprocess as sp
 import sys
 import time
 from collections.abc import Callable
+from ipaddress import AddressValueError, IPv6Address
 from typing import Any
 
 from selenium.webdriver.remote.remote_connection import urllib3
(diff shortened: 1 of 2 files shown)


lucasmariano003-wq:
@KazuCocoa, the three AppiumService fixes are ready for review: #1293 corrects the polling time unit, #1294 recognizes inline CLI arguments, and this PR formats IPv6 status URLs. Each is based independently on master, with local unit/type/style checks and HTTP or real Appium startup verification documented. Their fork workflows are awaiting maintainer approval. Could you review them and, if accepted, confirm the applicable compensation amount and any remaining claim requirements for each under the contributor scheme?

appium/webdriver/appium_service.py
```diff
@@ -315,7 +315,10 @@ def _make_status_path(args: list[str]) -> str:
 
 
 def _make_server_url(args: list[str]) -> str:
-    return f'{_parse_protocol(args)}://{_parse_host(args)}:{_parse_port(args)}{_make_status_path(args)}'
+    host = _parse_host(args)
+    if ':' in host and not host.startswith('['):
+        host = f'[{host}]'
+    return f'{_parse_protocol(args)}://{host}:{_parse_port(args)}{_make_status_path(args)}'
```
KazuCocoa, line 326: It makes sense to use IPv6Address checks below.
```
from ipaddress import AddressValueError, IPv6Address

host = _parse_host(args)
try:
    IPv6Address(host)
except AddressValueError:
    pass
else:
    host = f'[{host}]'
```

lucasmariano003-wq: Thanks for the suggestion. Updated `_make_server_url` to use `IPv6Address` and catch `AddressValueError`, as requested, in 5c7f15f. Added coverage for a scoped IPv6 address and an invalid IPv6-like value in f48b944; existing IPv4, hostname and already-bracketed cases remain covered. All 180 unit tests pass, along with Ruff, formatting and Mypy.

KazuCocoa (approved):

Issue #1147, opened 2025-07-23
mypy error "Incompatible return value type" on webdriver.find_element


bandophahita:
Mypy errors on the fact that `appium.webdriver.WebDriver.find_element` isn't overridden but indeed returns a different `WebElement`.

### Reproducible Code
```python
# no need to actually run this code, just send it through MyPy to see the error.

from appium.webdriver.common.appiumby import AppiumBy

from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from appium.webdriver import Remote
    from appium.webdriver.webelement import WebElement

def find_element(driver: Remote) -> WebElement:
    # should get a mypy error on next line
    return driver.find_element(AppiumBy.ACCESSIBILITY_ID, "demo button")


def find_elements(driver: Remote) -> WebElement:
    # should get a mypy error on next line
    return driver.find_elements(AppiumBy.ACCESSIBILITY_ID, "demo confetti")
```

### mypy output:
```
mypy .
screenpy_appium/target.py:89: error: Incompatible return value type (got "selenium.webdriver.remote.webelement.WebElement", expected "appium.webdriver.webelement.WebElement")  [return-value]
screenpy_appium/target.py:98: error: Incompatible return value type (got "list[selenium.webdriver.remote.webelement.WebElement]", expected "list[appium.webdriver.webelement.WebElement]")  [return-value]
Found 2 errors in 1 file (checked 35 source files)
```

### Solution
Annotations need to be added to `appium.webdriver.WebDriver`

```python
class WebDriver(
    webdriver.Remote,
    ...
):
    if TYPE_CHECKING:
        def find_element(self, by=By.ID, value: Optional[str] = None) -> MobileWebElement:
            ...
        def find_elements(self, by=By.ID, value: Optional[str] = None) -> list[MobileWebElement]:
            ...
```

KazuCocoa:
What about this?
https://github.com/appium/python-client/pull/1148

KazuCocoa:
Please try out `5.1.2`

KazuCocoa:
Hm, it looks like `Protocol` https://github.com/appium/python-client/blob/d43a190cc99dcae31f13c074b7f7ac9e7552f9d9/appium/protocols/webdriver/can_find_elements.py#L15-L24 needs some adjustment for mypy. I'll explore/learn a bit more

KazuCocoa:
https://github.com/appium/python-client/pull/1153

PR #1298, opened 2026-09-08
fix: preserve default locator strategy in WebDriver type hints

MattL1989:
The type-only `find_element` and `find_elements` declarations require `by`, although their inherited Selenium implementations default it to `By.ID`. This makes mypy reject valid calls such as `driver.find_element(value='target')` with a missing-argument error.

Add the inherited defaults and a consumer-level mypy regression covering omitted and explicit locator strategies while checking Appium `WebElement` return types. Runtime method dispatch is unchanged. This addresses a separate signature mismatch adjacent to #1147.

Validation on Windows with Python 3.12.14, Selenium 4.48.0 and mypy 2.3.1, using the repository's frozen development dependencies:

- The new regression fails before the fix with two missing-`by` errors and passes afterward.
- `python -m pytest test/unit/`: 187 passed (dependency deprecation and mocked-thread exception warnings emitted).
- `python -m mypy appium`: 320 source files pass.
- `python -m ruff check .` and `python -m ruff format --check .`: pass.

Prepared with AI assistance. No device-dependent functional tests were run; the correction affects static type declarations only.

Could you also confirm whether this correction qualifies for a contributor payment under the [Appium compensation scheme](https://github.com/appium/appium/blob/master/GOVERNANCE.md#compensation-scheme), and the applicable tier and expected payment timing? I am evaluating a small paid contribution trial and would like to confirm those details before taking on further work.


appium/webdriver/webdriver.py
```diff
@@ -20,6 +20,7 @@
     SessionNotCreatedException,
     WebDriverException,
 )
+from selenium.webdriver.common.by import By
```
KazuCocoa: I assume this import can be inside `if TYPE_CHECKING:`

```
if TYPE_CHECKING:
    from selenium.webdriver.common.by import By
```

KazuCocoa:
The amount would be adjusted, but basically the policy will work if it meed S and larger one. It will be in a couple of months usually.

test/unit/webdriver/type_hints_test.py
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env python
```
KazuCocoa: this test is not needed

appium/webdriver/webdriver.py
MattL1989: Valid, and safe. I checked the head commit: By appears only at line 23 (the import) and lines 285/288 and both stubs sit inside the class-level if TYPE_CHECKING: block, so By.ID is never evaluated at runtime.

test/unit/webdriver/type_hints_test.py
MattL1989: Understood — removed.

MattL1989:
Both points addressed in 26824aa — `By` import moved under `if TYPE_CHECKING:` and the test removed. Re-ran ruff, mypy and the unit suite locally (186 passed). Thanks also for clarifying the compensation tiering and timing.

KazuCocoa (approved):

PR #1289, opened 2026-09-05
test: add unit tests for orientation setter

Dor-bl:
## Description
Adds unit tests for the `orientation` getter and setter in the Appium Python Client to improve test coverage.

## Coverage
The new tests in `test/unit/webdriver/webdriver_test.py` cover:
- Getting the orientation (`test_orientation_getter`)
- Setting valid orientations (`'LANDSCAPE'` and `'PORTRAIT'`) (`test_orientation_setter`)
- Setting invalid orientations (e.g. `'INVALID'`), asserting that a `WebDriverException` is raised (`test_orientation_setter_invalid`)

PR #1307, opened 2026-09-15
test: add orientation getter and setter unit tests

KazuCocoa:
## Summary

Restore the intended orientation unit tests from #1289 following its revert in #1306. The existing orientation getter and setter already support the tested behavior.

Adds three tests in `test/unit/webdriver/webdriver_test.py`:
- Read `LANDSCAPE` from the orientation endpoint.
- Verify the request bodies when setting `LANDSCAPE` and `PORTRAIT`.
- Verify that an invalid orientation raises `WebDriverException` with the expected message.

The diff contains only these tests and their required imports (one file, 34 added lines). Assertions follow the current Ruff rules.

## Validation

- Full `test/unit/` suite passed using Python 3.14.6.
- All 24 tests in `test/unit/webdriver/webdriver_test.py` passed after the assertion style adjustment.
- Ruff lint and format checks passed for the changed file.
- `git diff --check` passed.

The test runs emitted httpretty deprecation warnings and thread warnings from the existing `test_session_catches_error` test.


KazuCocoa:
Cc @Dor-bl

PR #1309, opened 2026-09-20
fix(webelement): preserve default locator strategy in webelement type hints

Dor-bl:
### Summary

In Selenium's `WebElement.find_element` and `find_elements`, the locator strategy `by` parameter defaults to `By.ID` (`by: str = By.ID`).

While PR #1298 preserved the default `By.ID` locator strategy for `WebDriver`, `WebElement` and the `CanFindElements` protocol were omitted and continued to require `by` in their type hints (`by: str`). This causes static type checkers (such as `mypy`) to reject valid calls like `element.find_element(value='child')` or `element.find_elements(value='items')` with a missing positional argument error.

This PR adds the default `By.ID` locator strategy to `WebElement` and `CanFindElements`, completing the consistency established in #1298 across the client.

### Changes

- Added `by: str = By.ID` default to `find_element` and `find_elements` stubs in `appium/webdriver/webelement.py` (with `from selenium.webdriver.common.by import By` under `if TYPE_CHECKING:`).
- Added `by: str = By.ID` default to `appium/protocols/webdriver/can_find_elements.py`.
- Added unit tests in `test/unit/webdriver/webelement_test.py` verifying omitted `by` dispatch and protocol conformance.


appium/protocols/webdriver/can_find_elements.py
```diff
@@ -14,6 +14,8 @@
 
 from typing import TYPE_CHECKING, Protocol, runtime_checkable
 
+from selenium.webdriver.common.by import By
```
KazuCocoa, line 17: Move to under `if TYPE_CHECKING`?

